Issue #78, opened 2022-06-01
expose findings as prometheus metrics


josedonizetti:
create a prometheus exporter that exposes the results or trivy-operator as prometheus metrics.
For example, vulnerabilities as metrics.

TBD: spec how results map to metrics and labels

erikgb:
What kind of prometheus support are we looking for with this issue? Do we want:

1. Custom operator metrics, in addition to the ones provided by controller-runtime metrics?
2. Produce metrics from VulnerabilityReport, ConfigAuditReport CRDs? Ref. https://github.com/aquasecurity/trivy-operator/discussions/88
3. Both of the above?

IMO the answer is 3. The reason I ask, is that I might have an interesting contribution to this project if you think it might be a good idea. Somehow, we were not aware of this operator, nor it's predecessor starboard, until a few days ago. Still we have been using Trivy to scan images in our pipelines for quite some time. 😊

Scanning container images for workloads running in our clusters have been something we wanted to do for a long time. And struggling to find any good open source alternatives, that fulfilled our requirements, we started developing an in-house operator a couple of months ago. It works pretty well already, and we were supposed to put the operator into production this week. But due to some performance issues in the API towards our external image scanner (JFrog X-ray) it was postponed.

But then we discovered this operator! And (surprisingly) the design is somehow quite similar to what we have built - except for some obvious differences like internal/external image scanning. 👍 

Back to the issue service; metrics. We had a clear requirement for providing prometheus metrics for workload image scan results, and spent quite some time finding a fully working solution that scales and produces correct results. Our first approach was to add the metrics to the registry provided by controller-runtime to the operator. This solution had as least two issues:

1. The memory consumed by the controller container increased significantly because of the rich metrics stored in the internal registry.
2. Issues clearing out obsolete metrics without restarting the controller. The latter is still an [open issue](https://github.com/giantswarm/starboard-exporter/issues/48) in starboard-exporter, and pretty hard to solve - mainly because prometheus require complete matching labels to delete anything from the registry.

We managed to fix 2 by adding finalizers to our CRDs, but finalizers isn't something you want unless you really need it.

Looking for alternatives, we discussed providing the metrics from a separate workload, like starboard-exporter does. But it appeared less attractive, since this new workload would need to access the same API resources as already available in the operator cache.

[...]

PR #96, opened 2022-06-05
feat: add VulnerabilityReport summary metrics

erikgb:
## Description

This adds a leader-election aware custom Prometheus collector. This collector produces metrics from `VulnerabilityReport` summary, but can easily be extended to produce other metrics from trivy-operator custom resources.

## Related issues
- related to (but doesn't close) #78 

## Checklist

### DIFF
--- a/pkg/metrics/collector.go
+++ b/pkg/metrics/collector.go
@@ -1,0 +1,119 @@
+package metrics
+
+import (
+	"context"
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	k8smetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/trivy-operator/pkg/operator/etc"
+)
+
+var (
+	imageVulnLabels = []string{
+		"namespace",
+		"name",
+		"image_registry",
+		"image_repository",
+		"image_tag",
+		"image_digest",
+		"severity",
+	}
+	imageVulnDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("trivy", "vulnerabilityreport", "image_vulnerabilities"),
+		"Number of container image vulnerabilities",
+		imageVulnLabels,
+		nil,
+	)
+	severities = map[string]func(vs v1alpha1.VulnerabilitySummary) int{
+		"Critical": func(vs v1alpha1.VulnerabilitySummary) int {
+			return vs.CriticalCount
+		},
+		"High": func(vs v1alpha1.VulnerabilitySummary) int {
+			return vs.HighCount
+		},
+		"Medium": func(vs v1alpha1.VulnerabilitySummary) int {
+			return vs.MediumCount
+		},
+		"Low": func(vs v1alpha1.VulnerabilitySummary) int {
+			return vs.LowCount
+		},
+		"Unknown": func(vs v1alpha1.VulnerabilitySummary) int {
+			return vs.UnknownCount
+		},
+	}
+)
+
+type ResourcesMetricsCollector struct {
+	logr.Logger
+	etc.Config
+	client.Client
+}
+
+func (c *ResourcesMetricsCollector) SetupWithManager(mgr ctrl.Manager) error {
+	return mgr.Add(c)
+}
+
+func (c ResourcesMetricsCollector) Collect(metrics chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	targetNamespaces := c.Config.GetTargetNamespaces()
+	if len(targetNamespaces) == 0 {
+		targetNamespaces = append(targetNamespaces, "")
+	}
+	c.collectVulnerabilityReports(ctx, metrics, targetNamespaces)
+}
+
+func (c ResourcesMetricsCollector) collectVulnerabilityReports(ctx context.Context, metrics chan<- prometheus.Metric, targetNamespaces []string) {
+	vrList := &v1alpha1.VulnerabilityReportList{}
+	labelValues := make([]string, 7)
+	for _, n := range targetNamespaces {
+		if err := c.List(ctx, vrList, client.InNamespace(n)); err != nil {
+			c.Logger.Error(err, "failed to list vulnerabilityreports from API", "namespace", n)
+			continue
+		}
+		for _, vr := range vrList.Items {
+			labelValues[0] = vr.Namespace
+			labelValues[1] = vr.Name
+			labelValues[2] = vr.Report.Registry.Server
+			labelValues[3] = vr.Report.Artifact.Repository
+			labelValues[4] = vr.Report.Artifact.Tag
+			labelValues[5] = vr.Report.Artifact.Digest
+			for severity, countFn := range severities {
+				labelValues[6] = severity
+				count := countFn(vr.Report.Summary)
+				metrics <- prometheus.MustNewConstMetric(imageVulnDesc, prometheus.GaugeValue, float64(count), labelValues...)
+			}
+		}
+	}
+}
+
+func (c ResourcesMetricsCollector) Describe(descs chan<- *prometheus.Desc) {
+	descs <- imageVulnDesc
+}
+
+func (c ResourcesMetricsCollector) Start(ctx context.Context) error {
+	c.Logger.Info("Registering resources metrics collector")
+	if err := k8smetrics.Registry.Register(c); err != nil {
+		return err
+	}
+
+	// Block until the context is done.
+	<-ctx.Done()
+
+	c.Logger.Info("Unregistering resources metrics collector")
+	k8smetrics.Registry.Unregister(c)
+	return nil
+}
+
+func (c ResourcesMetricsCollector) NeedLeaderElection() bool {
+	return true
+}
+
+// Ensure ResourcesMetricsCollector is leader-election aware
+var _ manager.LeaderElectionRunnable = &ResourcesMetricsCollector{}
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aquasecurity/trivy-operator/pkg/configauditreport"
 	"github.com/aquasecurity/trivy-operator/pkg/ext"
 	"github.com/aquasecurity/trivy-operator/pkg/kube"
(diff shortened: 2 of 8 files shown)


erikgb:
Ok, I think I am ready for a first round of review! Test coverage could probably be higher, but the main parts are tested. This will also need some additions to the docs, but let's agree on the overall design, naming etc. before I try to add some words about this.

pkg/metrics/collector.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (c *ResourcesMetricsCollector) SetupWithManager(mgr ctrl.Manager) error {
+	return mgr.Add(c)
+}
+
+func (c ResourcesMetricsCollector) Collect(metrics chan<- prometheus.Metric) {
+	ctx := context.Background()
+	vrList := &v1alpha1.VulnerabilityReportList{}
+	if err := c.List(ctx, vrList, client.InNamespace("")); err != nil {
```
JAORMX, line 63: seems to me like this `List` call is dependent on the operator's configuration. While most folks will install it globally and this would apply for such cases. There are other configuration modes that might need to be taken into account. Perhaps the `ResourceMetricsCollector` initialization should also take a [Config](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/operator/etc/config.go#L12) object? that way, you'd be able to resolve the installation mode and determine the Namespaces to list VulnerabilityReporLists from.

For deployment modes such as the ones the Operator Lifecycle Manager uses, this is relevant, as the OLM itself determines RBAC permissions based on the target namespaces. And so, for other installation modes, the `List` call might fail.

JAORMX (changes requested):
Mostly this LGTM; I just left one comment about the List call that's used to gather the object to build the metrics from. Would also be nice to provide a ServiceMonitor as part of the operator deployment too.

pkg/metrics/collector.go
erikgb: Good catch! I forgot about the other install modes. I will fix this!

erikgb:
> Would also be nice to provide a ServiceMonitor as part of the operator deployment too.

I agree, but I think that is outside the scope of this PR. And installing a ServiceMonitor will require that the Prometheus operator (including its CRDs) are installed in the cluster.

JAORMX:
> > Would also be nice to provide a ServiceMonitor as part of the operator deployment too.
> 
> I agree, but I think that is outside the scope of this PR. And installing a ServiceMonitor will require that the Prometheus operator (including its CRDs) are installed in the cluster.

Fair point! I just thought it was a good idea to include a ServiceMonitor. But you're right, it could be added (as an example or as part of a `contrib/` directory) in another PR.

pkg/metrics/collector.go
erikgb: Target namespaces are now addressed in the PR. PTAL!

pkg/metrics/collector.go
```diff
@@ -0,0 +1,126 @@
...
+	k8smetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/trivy-operator/pkg/operator/etc"
+)
+
+var (
+	imageVulnLabels = []string{
+		"report_namespace",
+		"report_name",
```
erikgb, line 20: I am not sure how to name the namespace and name labels. The term `report_name` is borrowed from starboard-exporter, but I am not sure if it is correct in this context. Any input/comments?

pkg/metrics/collector.go
erikgb: The latest fix should assure this works with any Client, but I think it is not required when using the catching manager client, which I think should be used anyway....

pkg/metrics/collector.go
JAORMX: Well, the name of the metric already references the object. So I'd say that simply `name` and `namespace` are fine.

pkg/operator/operator.go
```diff
@@ -194,6 +195,17 @@ func Start(ctx context.Context, buildInfo trivyoperator.BuildInfo, operatorConfi
 			return fmt.Errorf("unable to setup clustercompliancereport reconciler: %w", err)
 		}
 	}
+
```
chen-keinan, line 198: I would make this configurable (can be default true), in case one wants to use an external metrics workload

chen-keinan (changes requested):

pkg/operator/operator.go
erikgb: I agree, and have added configuration (code) for this now. Docs needs to added/updated, but let's agree on the code first. I might need some pointers to how I should add this new feature to the docs...

erikgb:
@stone-z Thanks a lot for sharing your thoughts on this - I really appreciate your input from your work on starboard-exporter!

> I'm also dropping by to mention (hopefully constructively) that this approach is likely to have issues on larger clusters, if I understand the implementation correctly. Prometheus is going to hit scrape timeouts because reports will be fetched and processed on-demand. We have folks running the exporter in environments large enough for scrapes to time out even when the metrics are already cached. It's also possible (especially with multiple overlapping scrapes) that the memory needed to fetch reports and expose the metrics will put the pod over its memory limit. Throw in some network latency, priority/fairness, and multiple scrapes, and it's a bad day.

Why do you think this will add a considerable memory overhead? We are running with this approach in fairly large clusters without issues. On the contrary,  we actually had a lot of issues before changing _from_ the alternative approach you are suggesting - to produce metrics on reconcile. That seemed like the natural approach to begin with, but turned out to cause a lot of pain. Both the memory overhead of caching tons of metrics, and not to mention the inconsistent metrics long after restarting the operator.

Since the operator is already caching the relevant resources when using the controller-runtime `Client`, the resources should always be read from cache, and don't add any additional memory overhead to the operator. By producing metrics on demand (a `Collector`), I was hoping to avoid too much memory overhead by storing/caching the metrics in the internal Prometheus registry. I haven't looked too closely at the Prometheus internals, but I hope the implementation starts processing the metrics written to the channel, and not just aggregates the metrics.

> I'd also be happy to meet and discuss if anyone would find it valuable.

I would love to discuss this further with you! We could set up a Slack/Zoom call if we are able to find a suitable time slot. I am based in Oslo, Norway and currently on timezone GMT+2.

josedonizetti:
Thanks both @erikgb and @stone-z for the discussion here! Learning a lot. I've been studying and testing both approaches, would love to chat about it as well. I can make it any time that works for you two.

stone-z:
> Why do you think this will add a considerable memory overhead?

IIUC (and I might not, in fact, UC), metrics are not actually scraped until `Collect()` returns. In other words, each goroutine that adds a new metric might happen asynchronously, but the metrics register that gets built and scraped is the same, and thus has the same memory usage. The difference as I understand it would be that now there's the possibility of calling `Collect()` again before the previous scrape has finished, meaning an entirely new in-memory set of the same metrics. For a single scrape, I believe the memory requirements are the same, although the average utilization would be lower because metrics aren't stored between scrapes.

> inconsistent metrics long after restarting the operator

Side note - this is a side effect of starboard updating CRs which the exporter watches. I was able to use a finalizer for the deletion case, but due to the prometheus client limitations updates couldn't be handled. For starboard-exporter that will be fixed in the next release, but trivy-operator wouldn't have that problem as it is the one doing those problematic updates and can adjust the metrics itself.

> the operator is already caching the relevant resources

That's a good point. I'm still wrapping my head around tuning the cache a little bit as in some clusters we have lots of memory consumed by detail reports in the cache, but for purposes of summary metrics and because trivy-operator already caches report CRs you're right that there will be less of an issue with the roundtrips to the API server.

erikgb:
I think you might be correct about the metrics being collected and aggregated by the registry at scrape-time, so with non-overlapping scrapes the peak memory consumption would be about the same with both approaches. I am having a hard time finding docs in this area, but looking at the code it seems so - if I read it correctly.

> I was able to use a finalizer for the deletion case

We also tried that, but finalizers is an operational nightmare that IMO should be avoided unless you **really** need it.

I haven't tested the starboard/trivy operator at scale, but with our scanning-operator reconciling all resources after startup could take some time, up to an hour in some cases. Resulting in inconsistent metrics until all resources are reconciled at least once. Maybe that's not an issue here, but something to note.

@josedonizetti didn't have time for a call today, but we are hoping to be able to schedule a Zoom-call next week. Would you be interested in joining @stone-z? I think you have valuable input on this!

stone-z:
> Would you be interested in joining

Definitely :+1: where should we coordinate? I'm currently working from EDT time (GMT-4), available on Monday after ~15:30 CEST if Monday afternoon works for you both?

chen-keinan:
> > Would you be interested in joining
> 
> Definitely 👍 where should we coordinate? I'm currently working from EDT time (GMT-4), available on Monday after ~15:30 CEST if Monday afternoon works for you both?

@josedonizetti @JAORMX @erikgb I'll set up a call at 16:00 CEST time (let me know if anyone can't make it, I can reschedule)

stone-z:
@erikgb and I met and discussed this ([related Slack channel](https://app.slack.com/client/T02KSAWF2DT/C03K90PNXB7)). I am still of the opinion that this will have limitations under certain circumstances, but we agree that it's probably beyond the intended scope for Trivy's initial metrics support. 
We recommend trying this approach to provide basic summary metrics which will meet most users' needs. Users who need more can still run the `starboard-exporter` in parallel.

JAORMX (approved):

pkg/metrics/collector.go
```diff
@@ -0,0 +1,126 @@
...
+	client.Client
+}
+
+func (c *ResourcesMetricsCollector) SetupWithManager(mgr ctrl.Manager) error {
+	return mgr.Add(c)
+}
+
+func (c ResourcesMetricsCollector) Collect(metrics chan<- prometheus.Metric) {
+	vulnReports := make(chan v1alpha1.VulnerabilityReport)
+	go c.vulnerabilityReports(vulnReports)
```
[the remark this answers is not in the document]
erikgb: You are absolutely correct, Jose. I have removed the goroutine - at least for now. I think it should be ok for now, but please comment if you want me to adjust more. Code is trying to "keep the door open" for adding config metrics in a follow-up PR.

chen-keinan:
@erikgb supporting configauditreports CRD will be done in a separate PR?

erikgb:
> @erikgb supporting `configauditreports` CRD will be done in a separate PR? 

Yes, I think so. Wanted to discuss the approach in a simpler/smaller PR first. Which we have. Is that a problem? Planning to create the follow-up PR to that will close the referenced issue.

chen-keinan (approved):

josedonizetti (approved):
🎉

erikgb:
> But before I merge, can you please sign the commit? `git commit -s --amend`?

@josedonizetti Done! I had something wrong in my local Git setup.....

Issue #102, opened 2022-06-06
compliance: remove compliance CRD from deploy/static and helm


josedonizetti:
Perhaps we should not apply the CRDs for compliance on the cluster while we are not supporting it. It might be misleading.

![Screenshot from 2022-06-06 19-58-05](https://user-images.githubusercontent.com/34032/172262786-0f414ab1-f5b0-4468-8255-78cd09c7e10e.png)

josedonizetti:
@chen-keinan  wdyt?

chen-keinan:
ok, can be removed for now.

PR #109, opened 2022-06-07
temporarily remove compliance crds

josedonizetti:
- Close https://github.com/aquasecurity/trivy-operator/issues/102


chen-keinan:
@josedonizetti thinking about it again, maybe it is better to keep it with the partial result

josedonizetti:
@chen-keinan if we are keeping it, need to consider how to document, so users don’t think it is a bug, not to report on kube-bench checks, or don’t catch a problem because they think the test passed.

chen-keinan (approved):

Issue #111, opened 2022-06-07
trivy-operator: clustervulnerabilityreports only installed for helm


josedonizetti:
The CRD `ClusterVulnerabilityReport` is installed when using `helm` but not by the static `kubectl apply -f`.
Which one is correct? 

`helm install`
```
$ kubectl api-resources | grep aqua | grep vuln
clustervulnerabilityreports       clustervuln,clustervulns   aquasecurity.github.io/v1alpha1        false        ClusterVulnerabilityReport
vulnerabilityreports              vuln,vulns                 aquasecurity.github.io/v1alpha1        true         VulnerabilityReport
```

`kubectl apply -f`
```
$ kubectl api-resources | grep aqua | grep vuln
vulnerabilityreports              vuln,vulns           aquasecurity.github.io/v1alpha1        true         VulnerabilityReport
```

josedonizetti:
@chen-keinan do you know if we should have `clustervulnerabilityreports` installed by `kubectl apply -f` or should we not install it with `helm`?

chen-keinan:
@josedonizetti thinking about it again, maybe we should keep it with partial results

josedonizetti:
@chen-keinan partial results? Helm installing but kubeclt no?

chen-keinan:
> @chen-keinan partial results? Helm installing but kubeclt no?

any way we use to do kubectl manually

josedonizetti:
@chen-keinan Not following what you mean. Do you think is it okay for the two commands below to install something different? Shouldn't they be equivalent?

```
helm repo add aqua https://aquasecurity.github.io/helm-charts/
helm repo update
helm install trivy-operator aqua/trivy-operator \
   --namespace trivy-system \
   --create-namespace \
   --set="trivy.ignoreUnfixed=true" \
   --version v0.0.5
```

```
kubectl apply -f https://raw.githubusercontent.com/aquasecurity/trivy-operator/v0.0.5/deploy/static/trivy-operator.yaml
```

josedonizetti:
@chen-keinan and I spoke about it, and the `clustervulnerabilityreports` is legacy, we can remove it from the helm installation as well.

PR #113, opened 2022-06-08
remove clustervulnerabilityreports.crd.yaml

josedonizetti:
- Close https://github.com/aquasecurity/trivy-operator/issues/111


josedonizetti:
There is more related to this. I'll take a deeper look before making another PR.

chen-keinan:
@josedonizetti removing all the compliance-related will take an effort to bring it back, I suggest, maybe just changing the name or moving to temp folder, it will be easier to bring it back

josedonizetti:
@chen-keinan Is ‘clustervulnerabiltiyreport’ compliance? I don’t want to remove compliance code, we only disabled installing it! But from our conversation yesterday, I understood ‘clustervulnerabilityreport’ was legacy, no used anymore! Maybe misunderstood it.

PR #116, opened 2022-06-08
fix: remove compliance crd from itests

josedonizetti:


josedonizetti:
@chen-keinan merged without review because was a simple change to fix the main branch itests that were broken.

PR #186, opened 2022-06-24
Added Region support for ECR-Registry handling.

VF-mbrauer:
Added ECR-Region support as already done and working on old deprecated `starboard-operator`
Reference: https://github.com/aquasecurity/starboard/pull/1191/

### DIFF
--- a/pkg/plugin/trivy/plugin.go
+++ b/pkg/plugin/trivy/plugin.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/plugin/trivy/plugin.go
```diff
@@ -1486,3 +1499,10 @@ func getContainers(spec corev1.PodSpec) []corev1.Container {
 
 	return containers
 }
+
+func CheckAwsEcrPrivateRegistry(ImageUrl string) string {
```
josedonizetti, line 1503: nitpick: do we need to export this func?

josedonizetti (approved):
@VF-mbrauer Thank you.

PR #210, opened 2022-06-28
feat: add exposedsecrets metrics

josedonizetti:
Related to https://github.com/aquasecurity/trivy-operator/issues/78

### DIFF
--- a/pkg/metrics/collector.go
+++ b/pkg/metrics/collector.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+
 	"github.com/go-logr/logr"
 	ctrl "sigs.k8s.io/controller-runtime"
 
@@ -45,6 +46,35 @@ var (
 		},
 		"Unknown": func(vs v1alpha1.VulnerabilitySummary) int {
 			return vs.UnknownCount
+		},
+	}
+	exposedSecretLabels = []string{
+		"namespace",
+		"name",
+		"image_registry",
+		"image_repository",
+		"image_tag",
+		"image_digest",
+		"severity",
+	}
+	exposedSecretDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("trivy", "image", "exposedsecrets"),
+		"Number of image exposed secrets",
+		exposedSecretLabels,
+		nil,
+	)
+	exposedSecretSeverities = map[string]func(vs v1alpha1.ExposedSecretSummary) int{
+		"Critical": func(vs v1alpha1.ExposedSecretSummary) int {
+			return vs.CriticalCount
+		},
+		"High": func(vs v1alpha1.ExposedSecretSummary) int {
+			return vs.HighCount
+		},
+		"Medium": func(vs v1alpha1.ExposedSecretSummary) int {
+			return vs.MediumCount
+		},
+		"Low": func(vs v1alpha1.ExposedSecretSummary) int {
+			return vs.LowCount
 		},
 	}
 	configAuditLabels = []string{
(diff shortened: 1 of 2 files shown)


josedonizetti:
@erikgb ^^

pkg/metrics/collector.go
```diff
@@ -47,6 +48,35 @@ var (
 			return vs.UnknownCount
 		},
 	}
+	exposedSecretLabels = []string{
+		"namespace",
```
chen-keinan, line 52: probably not must for this PR , maybe we should make const from all the repeating fields name(namespace,name,severity and etc)

pkg/metrics/collector.go
```diff
@@ -138,6 +169,30 @@ func (c ResourcesMetricsCollector) collectVulnerabilityReports(ctx context.Conte
 	}
 }
 
+func (c ResourcesMetricsCollector) collectExposedSecretsReports(ctx context.Context, metrics chan<- prometheus.Metric, targetNamespaces []string) {
+	vrList := &v1alpha1.ExposedSecretReportList{}
```
erikgb, line 173:
```suggestion
	esrList := &v1alpha1.ExposedSecretReportList{}
```

Nit: Copy-pasted variable name. Maybe we should use more generic variable names, to make the code more copy-paste friendly? 😉 In this case `list` maybe?

pkg/metrics/collector.go
```diff
@@ -138,6 +169,30 @@ func (c ResourcesMetricsCollector) collectVulnerabilityReports(ctx context.Conte
...
+	labelValues := make([]string, 7)
+	for _, n := range targetNamespaces {
+		if err := c.List(ctx, vrList, client.InNamespace(n)); err != nil {
+			c.Logger.Error(err, "failed to list exposedsecretreports from API", "namespace", n)
+			continue
+		}
+		for _, vr := range vrList.Items {
```
erikgb, line 180:
```suggestion
		for _, esr := range vrList.Items {
```

Nit: Another copy-pasted variable name.

erikgb (changes requested):
A couple of nit comments, but generally LGTM.

pkg/metrics/collector.go
```diff
@@ -47,6 +48,35 @@ var (
...
+		"namespace",
+		"name",
+		"image_registry",
+		"image_repository",
+		"image_tag",
+		"image_digest",
+		"severity",
+	}
+	exposedSecretDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("trivy", "image", "exposedsecrets"),
```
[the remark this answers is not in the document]
josedonizetti: @chen-keinan Not sure about this. Exposed secrets will exist per image and also per resource. The way I see it we will have two different metric entries for it, like we would have for the resources themselves. No?

```
trivy", "image", "exposedsecrets -> for secrets inside images
trivy", "resource", "exposedsecrets -> for secrets inside resources
```

because although the same report, each report will be tied to a different resource, one for image, and another one for a k8s resoure.

pkg/metrics/collector.go
josedonizetti: yup, will create separate PR for it.

pkg/metrics/collector_test.go
```diff
@@ -1,9 +1,10 @@
 package metrics
 
 import (
-	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
 	"strings"
 
+	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
```
erikgb, line 6: Nice to fix the wrong placement of this import. I am used to 3 import "groups":

1. Standard Go imports
2. 3. party dependencies
3. Internal dependencies

WDYT? Do you have a standard for organizing imports in this project? Maybe for another PR?

pkg/metrics/collector.go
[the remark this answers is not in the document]
chen-keinan: Ok , then need probebly to think on a different naming for exposed data on resource , to avoid confusion

PR #211, opened 2022-06-28
docs: add integration/metrics

josedonizetti:
Related https://github.com/aquasecurity/trivy-operator/issues/78

### DIFF
--- a/docs/integrations/metrics.md
+++ b/docs/integrations/metrics.md
@@ -1,0 +1,3 @@
+# Metrics 
+
+`trivy-operator` exposed a `/metrics` endpoint by default  with metrics for vulnerabilities, exposed secrets, and configaudits.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -29,6 +29,7 @@ nav:
           - Overview: configuration-auditing/index.md
           - Built-in Configuration Audit Policies: configuration-auditing/built-in-policies.md
       - Integrations:
+          - Metrics: integrations/metrics.md
           - Lens Extension: integrations/lens.md
       - Tutorials:
           - Writing Custom Configuration Audit Policies: tutorials/writing-custom-configuration-audit-policies.md


josedonizetti:
@chen-keinan @erikgb This is just a place holder to say we have prometheus support. Let me know if you have suggestion of what to add, or we can improve it on a separate PR.

chen-keinan:
> @chen-keinan @erikgb This is just a place holder to say we have prometheus support. Let me know if you have a suggestion of what to add, or we can improve it on a separate PR.

@josedonizetti it's a good start, we should add to it more details around :
- pre-requisite 
- config: enable ServiceMonitor (after crds install)
- a reference to Prometheus install
- example of how metrics data look like  
- optional : Grafana configuration

mkdocs.yml
```diff
@@ -29,6 +29,7 @@ nav:
           - Overview: configuration-auditing/index.md
           - Built-in Configuration Audit Policies: configuration-auditing/built-in-policies.md
       - Integrations:
+          - Metrics: integrations/metrics.md
```
erikgb, line 32: I think metrics/observability represents more than an "integration", and would instead establish a new paragraph under "Docs" with the name "Observability". That will also give us more room for extending the docs over time, ref the comment from @chen-keinan.

docs/integrations/metrics.md
```diff
@@ -0,0 +1,3 @@
+# Metrics 
+
+`trivy-operator` exposed a `/metrics` endpoint by default  with metrics for vulnerabilities, exposed secrets, and configaudits.
```
erikgb, line 3: We must not forget that the metrics endpoint also expose interesting operator-metrics provided by controller-runtime.

chen-keinan (approved):

josedonizetti:
@erikgb I agree with your points, but let's discuss it on a separate issue PR, to unblock the release today, just a brief not will do it. Ok?

Issue #226, opened 2022-06-29
feat: add flag to disable exposedsecret scan


josedonizetti:
We can disable scans for configaudits and vulnerabilities. Let's add the same support to exposedsecrets.

PR #227, opened 2022-06-29
feat: add flag to disable secret scan

josedonizetti:

## Description

I chose a design for this feature that I would prefer to refactor later, mostly because of https://github.com/aquasecurity/trivy-operator/issues/415. Configurations to enable/disable scanner are considered `operatorConfig`, and are done through env variables, which are not passed in the context of the scanning plugin. Scanning configurations are configured as a ConfigMap, considered `trivyOperatorConfig`, and passed as context to the plugin. Because of this separation I'm setting the options from `operatorConfig` into `trivyOperatorConfig`.  Preferably I think we should only have one source of configuration, ConfigMap for all options related to the operator, which would be passed in the plugin context.

Note, I'm not talking about the specific scanner configmap (trivy-operator-trivy-config) which is yet another configuration file, specific only for the trivy plugin. 

## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/226

## Checklist

### DIFF
--- a/pkg/trivyoperator/config.go
+++ b/pkg/trivyoperator/config.go
@@ -51,6 +51,8 @@ type BuildInfo struct {
 type Scanner string
 
 const (
+	KeyVulnerabilityScannerEnabled       = "vulnerabilityScannerEnabled"
+	KeyExposedSecretsScannerEnabled      = "exposedSecretsScannerEnabled"
 	keyVulnerabilityReportsScanner       = "vulnerabilityReports.scanner"
 	KeyVulnerabilityScansInSameNamespace = "vulnerabilityReports.scanJobsInSameNamespace"
 	keyConfigAuditReportsScanner         = "configAuditReports.scanner"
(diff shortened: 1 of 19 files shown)


erikgb (changes requested):
Changes done LGTM, but shouldn't we also update the Helm chart (static resources) and config docs?

chen-keinan:
@josedonizetti maybe it better skipping the secret scan on trivy scan job command ,also good for performance :

`If your scanning is slow, please try '--security-checks vuln' to disable secret scanning`

pkg/trivyoperator/config.go
```diff
@@ -82,6 +84,30 @@ func GetDefaultConfig() ConfigData {
 	}
 }
 
+// Set sets a key on config data
+func (c ConfigData) Set(key, value string) {
```
chen-keinan, line 88: wonder if there is a way to support this change without adding `Set` method to `config.go` it always best to keep the config immutable (from code) after loading it

josedonizetti: @chen-keinan I also prefer it immutable. But I would prefer we merged this as is, and focused on refactoring this while also doing the task https://github.com/aquasecurity/trivy-operator/issues/415

Because having ConfigData immutable, but adding two keys (`KeyVulnerabilityScannerEnabled` and `KeyExposedSecretsScannerEnabled`) that are not part of the configmap, make the change more complex, and I didn't want to add them to `trivy-operator` configmap, because if we add it there will be two places to change flag (configmap and environment variable). 

Do you think releasing as is, and then I'll start on the config refactoring to fix this is okay?

chen-keinan: sure , Let's do it.

chen-keinan (approved):

Issue #204, opened 2022-06-28
Use controller-gen to generate code/configuration from Go markers


erikgb:
I notice that trivy-operator uses [code-generator](https://github.com/kubernetes/code-generator) to generate code for API types. While most of the controllers/operators I work with, both open-source and in-house, uses [controller-gen](https://github.com/kubernetes-sigs/controller-tools). As I see it, controller-gen seems better suited for developing/maintaining 3. party controllers/operators, while code-generator seems more for in-tree API/controllers.

Is there any particular reason for using code-generator over controller-gen? If not, I will suggest to migrate to controller-gen through issues/PRs. It will allow us to generate CRDs and controller-manager RBAC role from markers in the api/controller Go-code, in addition to the deep-copy funcs that also code-generator generates. It seems like the only feature that would get lost in such migration is the generated clientsets, informers and listers. But are these in use? It seems like they are not in use by the trivy-operator, and I think it is unlikely that other projects pull in this code as a dependency - since they will get issues with transitive dependencies to apimachinery etc.

The best available documentation for controller-gen I have found, is part of the kubebuilder book: https://book.kubebuilder.io/reference/markers.html.

If we decide to move forward with this, I would suggest a stepwise approach in the lines of:

1. Replace code-generator with controller-gen for generating the required deep-copy funcs. This will also include deleting all other generated code: https://github.com/aquasecurity/trivy-operator/tree/main/pkg/generated
2. Generate the trivy-operator CRDs from markers in code.
3. Generate the trivy-operator ClusterRole from markers in code. (Related issue https://github.com/aquasecurity/trivy-operator/issues/187)

chen-keinan:
@erikgb you can pick up this issue and raise a draft PR to be able to continue with evaluating the change

PR #232, opened 2022-07-01
Generate crd yaml files from go markers

padlar:
## Related issues
- Close #XXX

Remove this section if you don't have related PRs.

## Checklist
- [ ] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #233, opened 2022-07-01
rename the CRD files to match the controller-gen naming convention

padlar:
This patch is in preparation to switchover to the controller-gen for
generating CRD files from Go marker code. More importantly it makes it
easier to review the patch.

Related to #204


erikgb:
I think the compliance CRDs should be moved back to the crd folder, since I do not think we are able to control the CRD generation to that level of detail. Or do you have another plan for this @padlar? This means (partly) reverting https://github.com/aquasecurity/trivy-operator/pull/109. CC @josedonizetti

padlar:
> I think the compliance CRDs should be moved back to the crd folder (with the generate-names), since I do not think we are able to control the CRD generation to that level of detail. Or do you have another plan for this @padlar? This means (partly) reverting #109. CC @josedonizetti

this PR is about just renaming the CRD files. I was going to suggest moving the compliance CRDs back into the crd folder in the next PR as its not possible to generate the CRD files into 2 different folders automatically. Either manually move the compliance CRDs after generation or add a copy command post generation.

chen-keinan:
@padlar The compliance CRDs cannot be moved to crd folder at this point. 
while suggesting auto generation for crds , need to see how to overcome custom validation on crds which we do not wish to lose. #232

chen-keinan:
LGTM 🚀

chen-keinan (approved):

josedonizetti (approved):

Issue #187, opened 2022-06-24
Locking down Trivy operator RBAC


mtcolman:
Hi, I've just been reviewing the [rbac.yaml](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/static/02-trivy-operator.rbac.yaml) and I'm wondering if it's overly permissive given that a clusterRole and clusterRoleBinding are used?

I've seen `create`, `update` and `delete` actions are available on numerous resources and the binding is cluster wide (i.e. all namespaces), permission examples include:
  - Configmaps
  - secrets
  - serviceaccounts
  - Events
  - Jobs

Which I believe means I could copy secrets from one namespace to another and spin up jobs with malicious containers within. Could this instead be split into "get, list, watch" = all namespaces and "create, update, delete" namespaced to specific namespace it is required for? Or does it really need to be able to create the above in any/all namespaces in cluster?

Matt

erikgb:
@mtcolman Thanks for registering this issue! I agree that the ClusterRole assigned to the trivy-operator ServiceAccount has far too elevated permissions to cluster-wide resources, and should be reviewed and adjusted. I cannot image that the trivy-controllers actually need all these permissions, so I suspect this to be "leftovers" after removing `ClusterVulnerabilityReport` in https://github.com/aquasecurity/trivy-operator/pull/123. @josedonizetti @chen-keinan is this something you could look into?

chen-keinan:
@mtcolman  thank you for pointing out this concern.
I do see that we can improve permission grant to trivy-operator.

1. create a new Role (namespace scope)  for `create` , `update` and `delete` verbs and remove it from ClusterRole for the following
    - ConfigMaps
    - Secrets
    - ServiceAccount

2. keep  the following resources (create / update / delete )in ClusterRole as trivy-operator need it for it scanning capabilities
   - Jobs `(trivy-operator have a feature where it invoke a scan job on a different namespace )`
   
3. still to review if can changed grant on the permission can change
   - Events

erikgb:
I can work on this.

/assign

erikgb:
@chen-keinan @josedonizetti This is far more complicated than I imagined! My analyze so far, indicates that the current operator RBAC-setup isn't far from what is needed to support all features in trivy-operator for all operator setups (union of all requirements).

[...]

PR #234, opened 2022-07-01
fix: tighten RBAC; remove some unneeded cluster-wide permissions

erikgb:
## Description

This change withdraws some unneeded cluster-wide RBAC permissions and replaces them with equivalent operator namespace permissions - that are required to perform operator leader-election. To make it more obvious **why** permissions are needed, I put the new Role/RoleBinding in a dedicated file. This is also how kubebuilder scaffolds new projects. I can merge the new resources into the rbac.yaml file if you don't like it, but I prefer what I have suggested.

Technically the operator currently needs permission to create configmaps in the operator namespace, even with no leader election. Ref. https://github.com/aquasecurity/trivy-operator/blob/59452b4b465c1292aa9a763e459cf7f83553fb48/pkg/trivyoperator/plugin.go#L70-L85

But I decided to leave this out for now, since this permission is already covered by the new leader-election role. This can probably be improved when generating some of the RBAC, ref. https://github.com/aquasecurity/trivy-operator/issues/204

## Related issues
- Relates to #187, but doesn't fix it. 
- Relates to #204 

## Checklist

### DIFF
--- a/deploy/helm/templates/leader_election.yaml
+++ b/deploy/helm/templates/leader_election.yaml
@@ -1,0 +1,52 @@
+{{- if .Values.rbac.create}}
+---
+# permissions to do leader election.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "trivy-operator.fullname" . }}-leader-election
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "trivy-operator.labels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - list
+      - watch
+      - get
+      - create
+      - update
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - create
+      - get
+      - update
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "trivy-operator.fullname" . }}-leader-election
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "trivy-operator.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "trivy-operator.fullname" . }}-leader-election
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "trivy-operator.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end}}
--- a/deploy/static/02-trivy-operator.rbac.yaml
+++ b/deploy/static/02-trivy-operator.rbac.yaml
@@ -31,6 +31,7 @@ rules:
       - services
       - resourcequotas
       - limitranges
+      - configmaps
     verbs:
       - get
       - list
@@ -46,7 +47,6 @@ rules:
   - apiGroups:
       - ""
     resources:
-      - configmaps
       - secrets
       - serviceaccounts
     verbs:
@@ -61,12 +61,6 @@ rules:
       - secrets
     verbs:
       - delete
-  - apiGroups:
-      - ""
-    resources:
-      - events
-    verbs:
-      - create
   - apiGroups:
       - apps
     resources:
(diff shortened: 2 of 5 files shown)


deploy/static/02-trivy-operator.rbac.yaml
```diff
@@ -242,3 +228,68 @@ subjects:
...
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
chen-keinan, line 268: when `leases` was in `ClusterRole` it use to have only verbs (create, get, update) now in `Role` has additional (list, watch, patch, delete).
advice on this change

deploy/static/02-trivy-operator.rbac.yaml
```diff
@@ -242,3 +228,68 @@ subjects:
...
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
chen-keinan, line 256: why additional verbs (patch , delete) for configmap

deploy/helm/templates/leader_election.yaml
```diff
@@ -0,0 +1,59 @@
...
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
chen-keinan, line 35: when `leases` was in `ClusterRole` it use to have only verbs (create, get, update) now in `Role` has additional (list, watch, patch, delete).
advice on this change

deploy/helm/templates/leader_election.yaml
```diff
@@ -0,0 +1,59 @@
...
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
chen-keinan, line 23: why additional verbs (patch , delete) for configmap

chen-keinan (changes requested):

erikgb:
@chen-keinan There is a huge difference in cluster-wide permissions (ClusterRole/ClusterRoleBinding) compared to namespace permissions (Role/RoleBinding). The suggested leader-election role is based on the scaffolding performed if using kubebuilder: https://github.com/kubernetes-sigs/kubebuilder/blob/72f7c2ef3fd42d21756bc04029ecc499daee2533/pkg/plugins/golang/v2/scaffolds/internal/templates/config/rbac/leader_election_role.go#L43-L75. Since we do not control the leader election code (part of controller-runtime), I think it is better to give the operator full permissions on relevant resources in the operator namespace instead of reverse-engineering the code. The current default resource lock in c-r is `configmapsleases`, which will eventually change to `leases`.

chen-keinan:
@erikgb changing permission from cluster scope to namespace scope is well known and accepted change.
However the separation that was made in favour of `leading electron` has grant additional capabilities in namespace level to `trivy-operator` once you assigned it to `trivy-operator` service account.

from security perspective we did improved in cluster level however our security got worst in namespace level

erikgb:
> from security perspective we did improved in cluster level however our security got worst in namespace level

True, but overall this represents a security improvement, and I hope you agree on that. I do not know the leader election algorithm that well, but I would expect it to need most permissions on the relevant resources to do it's job in an efficient manner. Also I do not see the point of allowing i.e. the create and update verbs, but not patch verb - as patch is just another technique for updating the resource. When working with RBAC, I think it in general makes more sense to address permissions a bit more course-grained: read, write and read-write. By that I don't mean that more fine-grained RBAC should never be used....

chen-keinan:
>Also I do not see the point of allowing i.e. the create and update verbs, but not patch verb - as patch is just another technique for updating the resource. 
do you see `delete` permission necessary as well ? 
bear in mind that you grant permission to `trivy-operator` service account which can be used by malicious  pods

erikgb:
> bear in mind that you grant permission to trivy-operator service account which can be used by malicious pods

Well, I am a lot more worried about all the cluster-wide permissions granted to the trivy-operator service account. And I also think it makes sense to consider what others are doing faced with the exact same use case (kubebuilder). While I still disagree, I'll update the PR as requested to get this merged. 😄

deploy/static/02-trivy-operator.rbac.yaml
erikgb: Allowing same verbs on the namespace level now.

deploy/helm/templates/leader_election.yaml
erikgb: Allowing same verbs on the namespace level now.

deploy/helm/templates/leader_election.yaml
erikgb: Allowing same verbs on the namespace level now.

chen-keinan:
@erikgb thank you for this contribution it is an important one.

chen-keinan (approved):

PR #249, opened 2022-07-05
Remove support for deprecated configmaps leader election lock type

padlar:
"configmaps" hardcoded type is deprecated from v1.24+ and will be
removed from newer releases of client-go.

### DIFF
--- a/deploy/helm/templates/leader_election.yaml
+++ b/deploy/helm/templates/leader_election.yaml
@@ -10,16 +10,6 @@ metadata:
     {{- include "trivy-operator.labels" . | nindent 4 }}
 rules:
   - apiGroups:
-      - ""
-    resources:
-      - configmaps
-    verbs:
-      - list
-      - watch
-      - get
-      - create
-      - update
-  - apiGroups:
       - coordination.k8s.io
     resources:
       - leases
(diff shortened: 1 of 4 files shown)


deploy/helm/templates/leader_election.yaml
```diff
@@ -27,6 +17,10 @@ rules:
       - create
       - get
       - update
+      - list
+      - watch
+      - patch
+      - delete
```
chen-keinan, line 23: should not removed

deploy/static/02-trivy-operator.rbac.yaml
```diff
@@ -260,6 +249,10 @@ rules:
       - create
       - get
       - update
+      - list
+      - watch
+      - patch
+      - delete
```
chen-keinan, line 255: should be removed

deploy/static/trivy-operator.yaml
```diff
@@ -837,6 +827,10 @@ rules:
       - create
       - get
       - update
+      - list
+      - watch
+      - patch
+      - delete
```
chen-keinan, line 833: should be removed

PR #175, opened 2022-06-21
feat: add aggregated view clusterroles for default user-facing roles

erikgb:
## Description

Adds new aggregated view clusterroles for default user-facing roles. I have tested this on a multi-tenant Openshift cluster, and while I earlier was not allowed to see any trivy custom resources as a regular user, I can not view all resources:

image

## Related issues
- Close #173 

Remove this section if you don't have related PRs.

## Checklist
- [x] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #253, opened 2022-07-05
fix: add missing trivy-operator labels to user-facing roles

### LINKED ISSUE #173: Add clusterrole aggregating to default user-facing roles
User stories:

- As a standard user in a multi-tenant cluster, I would like to see the trivy custom resources in the namespace(s) I have access to.
- As a user with the cluster-reader (but not cluster-admin) cluster role, I would like to see all trivy custom resources in the cluster.

I think this can be done by adding a [aggregated clusterrole](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles), and I would like to submit a PR for this.

erikgb:
## Description

While analyzing trivy-operator RBAC, I noticed that the user-facing roles introduced in https://github.com/aquasecurity/trivy-operator/pull/175 are missing the labels applied to all trivy-operator resources. This PR fixes that.

## Related issues
- Close #173 

## Checklist

### DIFF
--- a/deploy/helm/templates/rbac.yaml
+++ b/deploy/helm/templates/rbac.yaml
@@ -170,6 +170,7 @@ kind: ClusterRole
 metadata:
   name: aggregate-config-audit-reports-view
   labels:
+    {{- include "trivy-operator.labels" . | nindent 4 }}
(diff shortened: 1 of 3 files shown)


josedonizetti (approved):
Thanks!

Issue #207, opened 2022-06-28
CI should verify that all generated is up-to-date


erikgb:
CI currently checks that the code generated by code-generator is up-to-date: https://github.com/aquasecurity/trivy-operator/blob/main/.github/workflows/build.yaml#L70-L71.

Since the static trivy-operator resources now are generated from the Helm chart, after https://github.com/aquasecurity/trivy-operator/pull/176 (and https://github.com/aquasecurity/trivy-operator/pull/185), I think it makes sense to include checks for that as well in our CI. (And also any other generated code/files that I might not be aware of...)

That will ensure a better contributor experience, since the CI for a PR will fail if this is forgotten by the PR author. We should ensure the error message is clear and informative, to lead the PR author to the corrective action.

Related PR, as an example on why we need this: https://github.com/aquasecurity/trivy-operator/pull/195

Also relevant issue: https://github.com/aquasecurity/trivy-operator/issues/204

chen-keinan:
When thinking about it again , I'm not sure we should commit the static files , probably better add them to .gitignore.
as it used only for testing and development , we can generate it ad hoc, as long as we have the script to do it correctly.
`testing` - during ci pipeline
`dev` - locally 

I believe it will simplified, maintenance , dev and release

erikgb:
> When thinking about it again , I'm not sure we should commit the static files , probably better add them to .gitignore. as it used only for testing and development , we can generate it ad hoc, as long as we have the script to do it correctly. `testing` - during ci pipeline `dev` - locally

Well, this is an interesting idea. IMO Helm is quite opinionated, and I know that a lot of users would prefer to avoid it. So personally I think we should officially support multiple options for installing trivy-operator. But I agree that supporting multiple installation approaches will require more from maintainers and contributors.

chen-keinan:
I assume that a users who want to use another method of deployment can always run the script (we need to keep the script updated) and auto generate these deployment files (should be also very cleared in documentation).

josedonizetti:
We could test the idea, support helm only and see what feedback we get, we can always return the statics.

badgerspoke:
> We could test the idea, support helm only and see what feedback we get, we can always return the statics.

Sorry to necropost (this seems to be the closest issue to my concern around static manifest), but can I ask why _some_ of the static manifests were removed but not all?

[...]

PR #319, opened 2022-07-13
chore: add .gitattributes with linguist-generated

erikgb:
## Description

This adds a `.gitattributes` file containing `linguist-generated` attributes making [GitHub keep certain files from displaying in diffs by default](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github).

Motivation: Ease the review process with focus on source files (by default) and improve the repository's language statistics.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+# Make GitHub keep certain files from displaying in diffs by default
+
+deploy/static/02-trivy-operator.rbac.yaml linguist-generated=true
+deploy/static/03-trivy-operator.config.yaml linguist-generated=true
+deploy/static/04-trivy-operator.policies.yaml linguist-generated=true
+deploy/static/05-trivy-operator.deployment.yaml linguist-generated=true
+deploy/static/06-trivy-operator.service.yaml linguist-generated=true
+deploy/static/trivy-operator.yaml linguist-generated=true


josedonizetti:
@erikgb I know the focus will be on the code, because those will be auto generated, but I kind of like to see those diffs. @chen-keinan what do you think?

erikgb:
> @erikgb I know the focus will be on the code, because those will be auto generated, but I kind of like to see those diffs. @chen-keinan what do you think?

Note: The diff will only not be displayed **by default**, and can always be expanded - as for large diffs. But I agree about the CRDs, and have now removed them from the list of linguist-generated files.

chen-keinan:
I don't mind having these files (static) as it is not important/has no affect on the product.

As mention in prev. issue #207 I do not think we should commit these files at all to git , as they do not contribute to the product or fail compilation and can be generated (by one simple command) at any point in time for local testing purposes.

Less files to review :)

erikgb:
> As mention in prev. issue #207 I do not think we should commit these files at all to git , as they do not contribute to the product or fail compilation and can be generated (by one simple command) at any point in time for local testing purposes.

I agree, but I don't think it is as simple as that - at least not without breaking things. But I'll close this for now, and instead focus on reducing the number of deduced/generated files.

PR #324, opened 2022-07-14
Prepare k8s 1.25 support

### LINKED ISSUE #221: Prepare support for K8s 1.25 APIs and removal of PSP
In order to prepare for the upcoming K8s 1.25 release, we should remove all use of already deprecated APIs that will be removed in 1.25. An overview of the APIs to be removed can be found here: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

I did a quick scan in the code-base, and it seems like we still use the following APIs:
- `batch/v1beta1 CronJob`; will need to be changed to use batch/v1 CronJob instead
- `policy/v1beta1 PodSecurityPolicy`; just remove use?

chen-keinan:

## Related issues
- Close #221 

## Checklist

### DIFF
--- a/pkg/kube/object.go
+++ b/pkg/kube/object.go
@@ -5,18 +5,18 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"github.com/aquasecurity/trivy-operator/pkg/trivyoperator"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
-	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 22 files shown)


pkg/kube/object.go
```diff
@@ -253,8 +256,66 @@ var ErrReplicaSetNotFound = errors.New("replicaset not found")
...
+	var resource client.Object
+	switch api {
+	case apiBatchV1beta1CronJob:
+		resource = &batchv1beta1.CronJob{}
+	case apiBatchV1CronJob:
+		resource = &batchv1.CronJob{}
+	default:
+		return nil, fmt.Errorf("api %s is not suooprted compatibale resource", api)
+	}
+	kindObjectMap[string(KindCronJob)] = resource
```
josedonizetti, line 308: This breaks the idea here for me, because we have `InitCompatibleMgr` doing a range on compatible types, and then calling `supportedObjectsByK8sKind` to return the map per `gvk`, but inside `supportedObjectsByK8sKind` we hardcode `resource` to a `KindCronJob`

pkg/kube/object.go
```diff
@@ -253,8 +256,66 @@ var ErrReplicaSetNotFound = errors.New("replicaset not found")
...
 type ObjectResolver struct {
 	client.Client
+	CompatibleMgr CompatibleMgr
+}
+
+func NewObjectResolver(c client.Client, cm CompatibleMgr) ObjectResolver {
+	return ObjectResolver{c, cm}
+}
+
+//InitCompatibleMgr initialized a CompatibleObjectMapper who store a map the of supported kinds with it compatible Objects (group/api/kind)
```
josedonizetti, line 278:
```suggestion
//InitCompatibleMgr initializes a CompatibleObjectMapper who store a map of the supported kinds with it compatible Objects (group/api/kind)
```

josedonizetti (commented):
@chen-keinan Looks good, two comments. Tested on 1.19/1.20/1.23/1.24, all seem good.

pkg/kube/object.go
chen-keinan: this the most flexible you can get as you need to provide [object type](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/configauditreport/controller/controller.go#L98) , but this whey we do not need to be depend on version , we know where is the computability problem so it can be managed

PR #306, opened 2022-07-12
feat: use kustomize to get labels back on generated resources

erikgb:
## Description

This PR suggests a light introduction of kustomize, without breaking changes, to reintroduce labels (lost by previous PRs) on generated resources.

## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/318
- Close https://github.com/aquasecurity/trivy-operator/pull/325

## Checklist
- [x] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #325, opened 2022-07-14
chore: remove some generated/deduced files

erikgb:
## Description

As noted by @chen-keinan in https://github.com/aquasecurity/trivy-operator/pull/319#issuecomment-1184038756 and https://github.com/aquasecurity/trivy-operator/issues/207#issuecomment-1169591044 some of the generated/deduced files do not have to committed to Git. This tries to remove as much as possible, but keeping `deploy/static/trivy-operator.yaml` - as it is mentioned in the docs as an installation alternative using kubectl.

We can also get rid of the trivy-operator.yaml if the static installation method is updated to use `kubectl` with kustomize. Which will make the new command

`kubectl create -k "https://github.com/aquasecurity/trivy-operator//deploy/static/?ref={{ git.tag }}"`. WDYT?

Note: I was unable to figure out how the integration tests were actually working before, but somehow they did. Works now also, and uses the same approach described in the contributor docs - which I think is nice.

## Checklist


hack/update-static.yaml.sh
```diff
@@ -12,24 +12,11 @@ trap "rm -rf $HELM_TMPDIR" EXIT
...
-cat $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/rbac.yaml \
-    $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/leader_election.yaml > $STATIC_DIR/02-trivy-operator.rbac.yaml
-cp $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/config.yaml $STATIC_DIR/03-trivy-operator.config.yaml
-cp $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/policies.yaml $STATIC_DIR/04-trivy-operator.policies.yaml
-cp $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/deployment.yaml $STATIC_DIR/05-trivy-operator.deployment.yaml
-cp $HELM_TMPDIR/trivy-operator-helm-template/trivy-operator/templates/service.yaml $STATIC_DIR/06-trivy-operator.service.yaml
+for i in `ls $CRD_DIR/*` $STATIC_DIR/namespace.yaml `ls $HELM_TMPDIR/trivy-operator/templates/*`
+do 
+  cat $i >> $HELM_TMPDIR/trivy-operator.yaml
+done
```
josedonizetti, line 20:
```suggestion
cat $CRD_DIR/* $STATIC_DIR/namespace.yaml $HELM_TMPDIR/trivy-operator/templates/* > $HELM_TMPDIR/trivy-operator.yaml
```

josedonizetti (commented):
@erikgb I like the idea of using kustomize and getting hid of trivy-operator.yaml as well, but perhaps in another PR? Maybe after wrapping this one up, you could fix https://github.com/aquasecurity/trivy-operator/pull/306 we have the labels back, before remove trivy-operator.yaml? WDYT?

chen-keinan:
@erikgb looks good! couple of comments I have I see that  the script `update-static.yaml.sh` do not generate the static files ( `rbac.yaml`, `config.yaml` and `policies.yaml`) , I usually use it for dev purposes , I do not use the `trivy-operator.yaml` because I do not want to deploy `trivy-operator' as well (as I run it from IDE). 

 I think it will be great to have it as part of the script and add it to `.gitignore` so we  keep it out of source code.
in addition I think we also need update the contribution guide , wdyt?

erikgb:
> I have I see that the script `update-static.yaml.sh` do not generate the static files ( `rbac.yaml`, `config.yaml` and `policies.yaml`) , I usually use it for dev purposes , I do not use the `trivy-operator.yaml` because I do not want to deploy `trivy-operator' as well (as I run it from IDE). I think it will be great to have it as part of the script and add it to `.gitignore` so we keep it out of source code.

@chen-keinan The use-case you describe is extremely easy to support with kustomize, ref. https://github.com/aquasecurity/trivy-operator/discussions/239. But we're not there yet, so I'll see what I can come up with.

>  in addition I think we also need update the contribution guide

What in particular do you think needs to be updated? I scanned through most of it, but couldn't find anything relevant to this.

hack/update-static.yaml.sh
erikgb: Thanks @josedonizetti! Fixed. I am not good at shell scripts! 😆

erikgb:
@chen-keinan PTAL! I've done a minor modification that I think will support your local workflow.

JAORMX (approved):

josedonizetti (approved):

erikgb:
@chen-keinan @josedonizetti The changes in this PR is now included in https://github.com/aquasecurity/trivy-operator/pull/306, so it's up to you if you want to merge this first and then https://github.com/aquasecurity/trivy-operator/pull/306 or just merge https://github.com/aquasecurity/trivy-operator/pull/306. 😄

Issue #237, opened 2022-07-03
vulnerabilityScannerScanOnlyCurrentRevisions does not delete old reports


chen-keinan:
**What steps did you take and what happened:**

* Enable vulnerabilityScannerScanOnlyCurrentRevisions on starboard operator
* See reports generated only for most current deployments
* Deploy new revision of deployment
* See report generated for new deployment
* See old report is still there

**What did you expect to happen:**

* old report being deleted as I only want reports for the current deployments

**Anything else you would like to add:**

* The implementation in #870 simply didn't add that part of the feature

**Environment:**

- Starboard version (use `starboard version`): `0.14.0`
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.2", GitCommit:"9d142434e3af351a628bffee3939e64c681afa4d", GitTreeState:"clean", BuildDate:"2022-01-19T17:27:51Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"clean", BuildDate:"2021-07-15T20:59:07Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
```

- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):
```
> sw_vers
ProductName:    macOS
ProductVersion: 11.6.2
BuildVersion:   20G314
```

erikgb:
I've had a look at this, and the issue seems to originate from https://github.com/aquasecurity/starboard/issues/926, which is referring to https://github.com/aquasecurity/starboard/pull/870. The title on https://github.com/aquasecurity/starboard/pull/870, "_feat(vulnrebilityreports) Only show the latest vulnrebilityreports from a deployment_", might fool you to believe that the change fixed this issue, but I don't think that ever was the plan. @NissesSenap, please correct me if I am wrong!

The name of the configuration property clearly states what is does, at least to me: vulnerabilityScanner**ScanOnly**CurrentRevisions. So I don't expect this to delete reports from inactive replicasets. There are also some interesting points in the discussion leading up to this change: https://github.com/aquasecurity/starboard/discussions/858.

[...]

PR #330, opened 2022-07-15
Unit test TTLReportReconciler controller

padlar:
Before making changes to the TTLReportReconciler, unit test the functionality

## Related issues
- related #237

### DIFF
--- a/pkg/vulnerabilityreport/ttl_report.go
+++ b/pkg/vulnerabilityreport/ttl_report.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 type TTLReportReconciler struct {
@@ -45,10 +46,15 @@ func (r *TTLReportReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *TTLReportReconciler) reconcileReport() reconcile.Func {
 	return func(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-		log := r.Logger.WithValues("report", req.NamespacedName)
+		return r.DeleteReportIfExpired(ctx, req.NamespacedName)
+	}
+}
+
+func (r *TTLReportReconciler) DeleteReportIfExpired(ctx context.Context, namespacedName types.NamespacedName) (ctrl.Result, error) {
+		log := r.Logger.WithValues("report", namespacedName)
 
 		report := &v1alpha1.VulnerabilityReport{}
-		err := r.Client.Get(ctx, req.NamespacedName, report)
+		err := r.Client.Get(ctx, namespacedName, report)
 		if err != nil {
 			if errors.IsNotFound(err) {
 				log.V(1).Info("Ignoring cached report that must have been deleted")
@@ -79,5 +85,4 @@ func (r *TTLReportReconciler) reconcileReport() reconcile.Func {
 		}
 		log.V(1).Info("RequeueAfter", "durationToTTLExpiration", durationToTTLExpiration)
 		return ctrl.Result{RequeueAfter: durationToTTLExpiration}, nil
-	}
 }
--- a/pkg/vulnerabilityreport/ttl_report_test.go
+++ b/pkg/vulnerabilityreport/ttl_report_test.go
@@ -1,0 +1,122 @@
+package vulnerabilityreport_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/trivy-operator/pkg/ext"
+	"github.com/aquasecurity/trivy-operator/pkg/operator/etc"
+	"github.com/aquasecurity/trivy-operator/pkg/trivyoperator"
+	"github.com/aquasecurity/trivy-operator/pkg/vulnerabilityreport"
+)
+
+const (
+	vreport = "vreport"
+	ns      = "default"
+)
+
+func TestRegenerateReportIfExpired(t *testing.T) {
+	// clock
+	clock := ext.NewSystemClock()
+
+	// scheme
+	scheme := trivyoperator.NewScheme()
+
+	// set the VulnerabilityScannerReportTTL
+	config, err := etc.GetOperatorConfig()
+	require.NoError(t, err)
+	hours, err := time.ParseDuration("24h")
+	require.NoError(t, err)
+	config.VulnerabilityScannerReportTTL = &hours
+
+	// logger object
+	logger := log.Log.WithName("testing")
+
+	// create TTLReport controller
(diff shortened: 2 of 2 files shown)


pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+		reportUpdateTimestamp time.Duration
+		ttlAnnotationMissing  bool
+		expectError           bool
+		reportDeleted         bool
+		ttlStr                string
+		invalidReportName     bool
+	}{
+		{
+			name:                  "Report timestamp < TTL",
+			reportUpdateTimestamp: -15 * 60 * time.Minute, // < 24h TTL
```
erikgb, line 57:
```suggestion
			reportUpdateTimestamp: -15 * time.Hour, // < 24h TTL
```

And also the others. Easier to read.

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+	// logger object
+	logger := log.Log.WithName("testing")
+
+	// create TTLReport controller
+	instance := vulnerabilityreport.TTLReportReconciler{Logger: logger, Config: config, Clock: clock}
+
+	tests := []struct {
+		name                  string
+		reportUpdateTimestamp time.Duration
+		ttlAnnotationMissing  bool
```
erikgb, line 49: I think you can get rid of this by treating the empty `ttlStr` as this case.

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+			ttlAnnotationMissing: true,
+			expectError:          false, // Ignoring report without TTL set
+		},
+		{
+			name:        "invalid TTL in the annotation",
+			ttlStr:      "badtime",
+			expectError: true,
+		},
+		{
+			name:              "invalid report name",
```
erikgb, line 80: I think this is more like non-existing report. And I don't think you need an extra property for it. Can't you just use the name?

erikgb (changes requested):

pkg/vulnerabilityreport/ttl_report_test.go
padlar: good catch! fixed it

pkg/vulnerabilityreport/ttl_report.go
```diff
@@ -2,8 +2,8 @@ package vulnerabilityreport
 
 import (
 	"context"
-	"fmt"
 	"time"
+	"fmt"
```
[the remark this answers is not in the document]
padlar: yes fixed it

pkg/vulnerabilityreport/ttl_report_test.go
padlar: this flag removes the entire annotation when set, so helpful in testing missing entire annotation and invalid duration value

pkg/vulnerabilityreport/ttl_report_test.go
padlar: yes fixed the name. extra property is required to set the name to an empty string.

pkg/vulnerabilityreport/ttl_report_test.go
erikgb: I think you can use the same conditional with `ttlStr != ""`.

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+	// logger object
+	logger := log.Log.WithName("testing")
+
+	// create TTLReport controller
+	instance := vulnerabilityreport.TTLReportReconciler{Logger: logger, Config: config, Clock: clock}
+
+	tests := []struct {
+		name                  string
+		reportUpdateTimestamp time.Duration
+		expectError           bool
```
erikgb, line 49:
```suggestion
		wantErr           bool
```

Nit

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+		reportDeleted         bool
```
erikgb, line 50:
```suggestion
		wantReportDeleted         bool
```

Nit

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,124 @@
...
+		ttlStr                string
+		invalidReportName     bool
+		ttlAnnotationMissing  bool
+	}{
+		{
+			name:                  "Report timestamp < TTL",
+			reportUpdateTimestamp: -15 * time.Hour, // < 24h TTL
+			expectError:           false,
```
erikgb, line 58: False is the zero value of bool, so I would prefer if you don't set it in those cases. To make the test more compact an easier to read. Applies to all similar cases.

erikgb (changes requested):

pkg/vulnerabilityreport/ttl_report_test.go
padlar: removed

pkg/vulnerabilityreport/ttl_report_test.go
padlar: fixed

pkg/vulnerabilityreport/ttl_report_test.go
padlar: updated

pkg/vulnerabilityreport/ttl_report_test.go
padlar: updated the code to explicitly set the name

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,126 @@
...
+		wantError             bool
+		wantReportDeleted     bool
+		ttlStr                string
+		invalidReportName     bool
+	}{
+		{
+			name:                  "Report timestamp < TTL",
+			reportUpdateTimestamp: -15 * time.Hour, // < 24h TTL
+			wantError:             false,
+			wantReportDeleted:     false,
```
erikgb, line 63:
```suggestion
```

In this case I do not think this adds any value.

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,126 @@
...
+			ttlStr:                "24h",
+		},
+		{
+			name:                  "Report timestamp exceeds TTL",
+			reportUpdateTimestamp: -25 * time.Hour, // > 24 TTL
+			wantError:             false,
```
erikgb, line 69:
```suggestion
```

Adds no value

pkg/vulnerabilityreport/ttl_report_test.go
```diff
@@ -0,0 +1,126 @@
...
+		{
+			name:      "invalid TTL in the annotation",
+			ttlStr:    "badtime",
+			wantError: true,
+		},
+		{
+			name:              "non-existent report name",
+			invalidReportName: true,  // sets the report name to empty string
+			wantError:         false, // missing/invalid report ignored
+			wantReportDeleted: false,
```
erikgb, line 87:
```suggestion
```

No meaning in expecting a delete of something that doesn't exist.

erikgb (changes requested):
Almost LGTM, but I am suggesting to remove some zero-values from the test fixtures. But keeping the cases where we want to express the intent.

pkg/vulnerabilityreport/ttl_report_test.go
padlar: removed

pkg/vulnerabilityreport/ttl_report_test.go
padlar: removed

josedonizetti (approved):
Thank you!

Issue #335, opened 2022-07-16
Testing controllers with envtest cluster


erikgb:
Working with trivy-operator for a couple of weeks, I already managed to propose changes that introduced bugs. And these bugs were not detected by CI and merged into main. That tells me we need better tests. 😄

I am unable to run integration tests locally for two reasons:

1. My development environment is docker-based, and kind does not run on docker (dind). I've been using [k3d](https://k3d.io/) as a workaround, which can run on docker as an alternative, but there is always a risk when having to modify local files to run things.
2. Because I work in an air-gapped environment, trivy (scanner) is unable to update the DB without modifications to trivy-operator configuration.

To mend this, I am suggesting to introduce envtest clusters to run (integration) tests. How this might look like, is fairly well documented in the [kubebuilder book](https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html). This will allow the integration tests to run in a standard Go test - without bootstrapping a full cluster (like kind).

- Running tests using `make` will be supported without requirements to install anything, as there is a [nice tool to setup envtest](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest) available. Running tests for multiple versions of Kubernetes is also possible with ease.
- For running tests from a local development environment (IDE) you would have to install kubebuilder-tools, which is a small package containing kubectl, etcd and api-server.

I can prepare a PR suggesting envtest integration tests for vulnerability scanning, so you can have a look at the details in this proposal. We (as in my team/company) are currently only interested in the vulnerability scanning feature in trivy-operator, so I am not prepared to do the work on the other controllers. But if the community likes this idea, I hope that someone else can contribute writing tests for the remaining controllers!

If you think this approach looks good, I think this could be an overall plan for refactoring/improving the tests in trivy-operator:

1. Write envtest cluster tests for all controllers in trivy-operator. Since we already have unit tests (that will not change), I think these tests should be fairly simple, at least in the beginning.
2. Introduce e2e (end-to-end) tests to run tests against a real cluster (kind). The goal for the initial version of these tests will be to be good enough to delete the existing integration tests (next task). We (my team) are using [kuttl](https://kuttl.dev/) to drive internal operator e2e-tests, but I have no strong feelings about tooling here. It works well for us, and allows us to write good tests with little effort. We can decide on the tooling later.
3. Delete the existing integration tests, which I think are confusing and hard to use/improve.

chen-keinan:
Great initiative looking forward to the PR

[...]

PR #337, opened 2022-07-16
test: add vulnerability scan job creation integration tests

erikgb:
## Description

Introducing envtest cluster integration tests that can run without bootstrapping a full K8s cluster. This is the first step towards replacing the existing kind-based integration tests with envtest integration tests + end-to-end (e2e) tests using kind as proposed in #335.

This first PR adds integration tests for vulnerability scan job creation, and will be followed up by a new PR containing similar tests for finished scan job reconciling.

I will also create a PR for e2e-testing using [kuttl](https://kuttl.dev/), which together with the new integration tests should allow us to delete the existing integration tests. At least it seems so to me - since they appear quite "thin" and full of TODOs.

The contributor docs might appear a bit confusing in the transition period, since we now will have two types of "integration tests", but I hope that is acceptable for a short period.

## Related issues
- Relates to #335, but does not close it

## Checklist


erikgb:
@chen-keinan Maybe you can scan through this draft and tell me what you think before I add tests for more workload types. 🙏

chen-keinan:
> @chen-keinan Maybe you can scan through this draft and tell me what you think before I add tests for more workload types. 🙏

the test case look ok , isn't the envtest nfra is the same we have today : [envtest](https://github.com/aquasecurity/trivy-operator/blob/main/itest/trivy-operator)?

erikgb:
> isn't the envtest nfra is the same we have today : [envtest](https://github.com/aquasecurity/trivy-operator/blob/main/itest/trivy-operator)?

WDYM? The itests run with kind. Envtest spins up a simple envtest cluster (from within the test suite), with just etcd and api-server - nothing more, nothing less. Did you try it? It works very well IMO.

chen-keinan:
> > isn't the envtest nfra is the same we have today : [envtest](https://github.com/aquasecurity/trivy-operator/blob/main/itest/trivy-operator)?
> 
> WDYM? The itests run with kind. Envtest spins up a simple envtest cluster (from within the test suite), with just etcd and api-server - nothing more, nothing less. Did you try it? It works very well IMO.

does it require `Kubebuilder` bin to run locally ?

erikgb:
> > > isn't the envtest nfra is the same we have today : [envtest](https://github.com/aquasecurity/trivy-operator/blob/main/itest/trivy-operator)?
> > 
> > 
> > WDYM? The itests run with kind. Envtest spins up a simple envtest cluster (from within the test suite), with just etcd and api-server - nothing more, nothing less. Did you try it? It works very well IMO.
> 
> does it require `Kubebuilder` bin to run locally ?
> 
> ```
> ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 0, "error": "fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
> ```

No requirement for the kubebuilder CLI, but you need to install envtest for running tests from a IDE or standalone: https://book.kubebuilder.io/reference/envtest.html#installation. The Makefile takes care of this to run tests in CI (or locally). Does `make test` work for you?

chen-keinan:
> Does `make test` work for you?

yes it works and I like it.

erikgb:
> couple of comments , need to check if the current tests in `itests can be migrated` and I assume you plan to put it under folder on the main project

I am not sure if I understand what you re writing here. 🤔 My suggested overall plan is described in https://github.com/aquasecurity/trivy-operator/issues/335 (last part).

pkg/vulnerabilityreport/suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
+
+func loadResource(obj runtime.Object, filename string) error {
+	yamlFile, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+	err = yaml.UnmarshalStrict(yamlFile, obj)
```
josedonizetti, line 135:
```suggestion
	return yaml.UnmarshalStrict(yamlFile, obj)
```

pkg/vulnerabilityreport/workload_controller_test.go
```diff
@@ -0,0 +1,72 @@
...
+			createdJob := &batchv1.Job{}
+
+			// We'll need to retry getting this newly created Job, given that creation may not immediately happen.
+			Eventually(func() error {
+				return k8sClient.Get(ctx, jobLookupKey, createdJob)
+			}, timeout, interval).Should(Succeed())
+
+			Expect(createdJob).Should(WithTransform(NormalizeUntestableScanJobFields, Dyqual(expectedJob)))
+		},
+		Entry("Should create a scan Job for CronJob", &batchv1.CronJob{}, "cronjob.yaml", "cronjob-scan-job.yaml"),
```
josedonizetti, line 64: @erikgb Can we rename the expected files? Maybe `expected-cronjob-scan.yaml`, at least for me it was not very intuitive to look into the `testdata` folder and see difference.

josedonizetti (commented):
@erikgb This looks great. Thank you! Two minor comments.

pkg/vulnerabilityreport/workload_controller_test.go
erikgb: I agree, but can we rather name them `<resource>-expected-scan.yaml`? I like to keep them next to the resource-file.

pkg/vulnerabilityreport/suite_test.go
erikgb: Fixed, thanks

pkg/vulnerabilityreport/workload_controller_test.go
josedonizetti: 👍

erikgb: Renamed

chen-keinan:
LGTM , in term of folder structure I would suggest to move it one level up , to be in project level, so we could extent the itest for other scanners and reuse suite code

erikgb:
> I see this almost as a controller unit tests with mocks, e2e would be more close to what we have currently on itests.

I do not agree with you here. 😉  Not using any mocks in these tests, so they are definitely integration tests (of the controller)! I'll probably mock out the logs reader in my next PR to add integration tests for the scan job controller. But that is because I have to, since api-server nor etcd can provide me with pod logs.

I would also like to fix https://github.com/aquasecurity/trivy-operator/issues/370 before adding integration tests for the scan job controller, but for now I'll just have to wait for your reviews and merge. What's holding this up? Just time to review/merge, or is there something else? Please tell me! 😺

chen-keinan:
> > in term of folder structure I would suggest to move it one level up , to be in project level, so we could extent the itest for other scanners and reuse suite code
> 
> If I don't misunderstand you, I am against this. We should ideally have one integration test per controller, each of them getting a separate envtest cluster. Reusing code is of course something we should do! 

Not sure I follow , there is a lot of `testdata/fixture` and some of `suite_test` code that can be reused why duplicate ?

erikgb:
@josedonizetti I am not sure if I am able to figure out what causes the race, but this PR is introducing ansync testing, which is new in trivy-operator. So I suspect it to already be an issue in the code on main, and that this PR only detects it. But I have never seen this race, nor locally on my workstation or in CI - and I have rebased numberous times triggering a new CI run.

But looking at the stack-traces, I think this piece of code looks suspicious:

https://github.com/aquasecurity/trivy-operator/blob/main/pkg/vulnerabilityreport/controller.go#L79-L80

Setting up two watches against the same GR seems like a bad idea. It should be either `v1.CronJob` or `v1beta1.CronJob` right?

erikgb:
> 
> Not sure I follow , there is a lot of `testdata/fixture` and some of `suite_test` code that can be reused why duplicate ?

I suggest to reuse when we can reuse. The bootstrapping can be reused, but that should be part of the production code and not the test code. Not sure if we will/should reuse the resources.

erikgb:
> > > Not sure I follow , there is a lot of `testdata/fixture` and some of `suite_test` code that can be reused why duplicate ?
> > 
> > 
> > I suggest to reuse when we can reuse. The bootstrapping can be reused, but that should be part of the production code and not the test code. Not sure if we will/should reuse the resources.
> 
> not sure I understand , most of code in `suite_test` can be reused , do you agree on it ?

No

josedonizetti:
> Setting up two watches against the same GR seems like a bad idea. It should be either `v1.CronJob` or `v1beta1.CronJob` right?

Fixing this did the trick, no data races happening anymore. :D

PR #276, opened 2022-07-10
WIP: chore: disable caching of secrets and serviceaccounts

erikgb:
## Description

Disables caching of secrets and serviceaccounts to avoid overloading the API server and tighten operator RBAC permissions.

## Related issues
- Close #XXX

Remove this section if you don't have related PRs.

## Checklist
- [ ] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

Issue #230, opened 2022-06-30
On a large Cluster - Timeout to list secrets


cdesaintleger:
**What steps did you take and what happened:**

Hello, 
on a large cluster (250+ worker nodes , ~20k secrets) the operator return a lot of error messages about "secret listing timeout" (trace below ).
I think that , the operator query the dataplane to get all secrets in all namespaces, and reach a timeout 60sec. 

I don't understand why the operator want to get all secrets in one time. 
Maybe there is an option to tune it. 

`Trace[202191422]: [1m4.391699449s] [1m4.391699449s] END
E0630 13:10:31.239778       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.1/tools/cache/reflector.go:167: . Please retry. Original error: stream error: stream ID 37085; INTERNAL_ERROR; received from peer
W0630 13:11:33.944659       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.1/tools/cache/reflector.go:167: failed to list *v1.Secret: stream error when reading response body, may be caused by closed connection. Please retry. Original error: stream error: stream ID 37089; INTERNAL_ERROR; received from peer
I0630 13:11:33.944735       1 trace.go:205] Trace[1291045998]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.24.1/tools/cache/reflector.go:167 (30-Jun-2022 13:10:33.624) (total time: 60319ms):`

**Anything else you would like to add:**

GKE 1.21
250+ nodes
20K secrets 

**Environment:**

- Trivy-Operator version 0.0.7
- Kubernetes version 1.21

Thank you for your assistance :)
Ch

chen-keinan:
@cdesaintleger following to reviewing  this issue , I can confirm the following :
- trivy-operator doesn't use `client list secrets` as part of it functionality , in some case we get a single secret with  : 
   `r.client.Get(ctx, client.ObjectKey{Name: secretRef.Name, Namespace: ns}` with specific refName and namespace but never use ***secrets listing***.
   
[...]

PR #338, opened 2022-07-17
Draft: refactor: use manager APIReader instead of disabling cache

erikgb:
## Description

After reaching out to controller-runtime maintainers on Slack, I got a [nice suggestion](https://kubernetes.slack.com/archives/C02MRBMN00Z/p1657939191076559?thread_ts=1657889202.276989&cid=C02MRBMN00Z) to use `mgr.GetAPIReader()` instead of having to disable cache for secrets and serviceaccounts. And I think that is an improvement.

Current status: This is still a draft PR, since my local testing shows that with this change, the operator is again requesting access to list secrets cluster-wide - which we don't want.

## Related issues
- Relates to https://github.com/aquasecurity/trivy-operator/pull/276
- Relates to https://github.com/aquasecurity/trivy-operator/issues/230

## Checklist

### DIFF
--- a/pkg/kube/secrets.go
+++ b/pkg/kube/secrets.go
@@ -115,15 +115,15 @@ type SecretsReader interface {
 // NewSecretsReader constructs a new SecretsReader which is using the client
 // package provided by the controller-runtime libraries for interacting with
 // the Kubernetes API server.
-func NewSecretsReader(c client.Client) SecretsReader {
-	return &secretsReader{client: c}
+func NewSecretsReader(r client.Reader) SecretsReader {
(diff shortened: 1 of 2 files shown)


chen-keinan:
@erikgb why the didn't recommended to disable cache for secrets and serviceaccounts. ?

erikgb:
> @erikgb why the didn't recommended to disable cache for secrets and serviceaccounts. ?

@chen-keinan It will work that also, so I think we should leave it as is now. I've been looking at how to refactor this to use the APIReader, but it will require some work. Closing this.

Issue #121, opened 2022-06-10
more granular way to set up security context in helm chart


3rgfbrgf:
It would be nice to have a way to set up securityContexts for:
 - initContainers
 - jobs, which are being run during scans. Including compliance scan jobs, which require a lot of host-level access.

chen-keinan:
@3rgfbrgf could you add an example ?

3rgfbrgf:
@chen-keinan sure.
Let's take as an example any scan-vulnerabilityreport job spec.
There is an initContainer with an empty securityContext.
We have an admission controller in our cluster, which prevents this from running by design.

As for compliance scans - pod for that job should mount host paths and so on.
It should be configurable by providing clear and obvious permissions for host PID, host network, host path etc in security context.
Currently it just needs privileged mode, as I can understand.

Also, PSP is deprecated, but still the case, so it's better to have an opportunity to create and customise it in your helm chart.

chen-keinan:
@3rgfbrgf please confirm if the PR #362 solve your use case or you need it more granular in container level

chen-keinan:
closing this issue as it been implemented by #362 @3rgfbrgf if still an issue feel free to open it

FelixKampfer:
Hi, I thinkk #362 doesn't address the securityContext of initContainers. I opened https://github.com/aquasecurity/trivy-operator/pull/459 as a possible solution.

PR #362, opened 2022-07-19
feat: setup securityContext on plugins

nerzhul:
At Veepee we only permit pods to run as non root. Plugin works properly as non root, add variables to setup it as non root and permit to customize UID

## Description

Ref #121 
## Checklist


nerzhul:
I just fixed tests, it's now mergeable, if you can review and release when you can, we'd glad to do more test on our 200 nodes cluster :)

docs/operator/configuration.md
```diff
@@ -50,6 +50,7 @@ To change the target namespace from all namespaces to the `default` namespace ed
 | `scanJob.tolerations`| N/A| JSON representation of the [tolerations] to be applied to the scanner pods so that they can run on nodes with matching taints. Example: `'[{"key":"key1", "operator":"Equal", "value":"value1", "effect":"NoSchedule"}]'`|
 | `scanJob.annotations`| N/A| One-line comma-separated representation of the annotations which the user wants the scanner pods to be annotated with. Example: `foo=bar,env=stage` will annotate the scanner pods with the annotations `foo: bar` and `env: stage` |
 | `scanJob.templateLabel`| N/A| One-line comma-separated representation of the template labels which the user wants the scanner pods to be labeled with. Example: `foo=bar,env=stage` will labeled the scanner pods with the labels `foo: bar` and `env: stage`|
+| `scanJob.podSecurityContext`| N/A| One-line JSON representation of the template securityContext which the user wants the scanner pods to be secured with. Example: `{"RunAsUser": 1000, "RunAsGroup": 1000, "RunAsNonRoot": true}`|
```
erikgb, line 53: Will only "One-line JSON representation" be supported by this? I think it would be nice to support YAML-format - if possible, as this will be a better UX IMO. But probably not possible?

pkg/vulnerabilityreport/builder_test.go
```diff
@@ -207,7 +209,9 @@ func TestScanJobBuilder(t *testing.T) {
 							trivyoperator.LabelResourceSpecHash:           "7d6c657bf7",
 						},
 					},
-					Spec: corev1.PodSpec{},
+					Spec: corev1.PodSpec{
+						SecurityContext: &corev1.PodSecurityContext{},
```
erikgb, line 213: I don't think this test should change, if I don't misunderstand the test. As this new feature should not change the existing job spec, if not using this feature. Multiple occurrences.

pkg/trivyoperator/config.go
```diff
@@ -56,6 +56,7 @@ const (
 	keyConfigAuditReportsScanner         = "configAuditReports.scanner"
 	keyScanJobTolerations                = "scanJob.tolerations"
 	keyScanJobAnnotations                = "scanJob.annotations"
+	keyScanJobPodSecurityContext         = "scanJob.podSecurityContext"
```
erikgb, line 59:
```suggestion
	keyScanJobPodSecurityContext         = "scanJob.podTemplateSecurityContext"
```

To be better aligned with the existing property `scanJob.podTemplateLabels`?

pkg/vulnerabilityreport/builder_test.go
nerzhul: ok you prefer to keep previous behaviour, and not add any empty spec, let's look at this

erikgb:
@nerzhul Thanks for this PR! I have added a few comments. We run on Openshift, and it is really picky about securityContext. It has it's SCC that want to control all that.

docs/operator/configuration.md
erikgb: Isn't this configured from a configmap? But I see that the similar property `scanJob.templateLabel` (also a typo; should be Label**s**) is using a simple string. I'll leave this decision up to the maintainers....

nerzhul: it's the configmap key exact, i use JSON because GetScanJobTolerations uses also a JSON to be consistent

nerzhul:
@erikgb i addressed your concerns, except the one on the format, let's wait for maintainers, but it kept consistency with tolerations, both are JSON unmarshaled.

For Openshift, you have specific things. On our side we use PSP and PodSecurities on all namespaces, then it's not possible to run root pods. With this patch pods runs perfectly as UID > 1000 users and creates reports, on a vanilla Kubernetes with just PSP/PS

erikgb:
> For Openshift, you have specific things. 

I am not against adding this new feature, but I don't think it should break the existing operator. :-)

nerzhul:
no problem @erikgb , now it's more close to the original one, regarding the tests on other parts :) 
Tests on our dev cluster shows it's fine.
I just amened the helm chart to make helm chart use YAML syntax and let it transform to JSON like tolerations

pkg/trivyoperator/config.go
```diff
@@ -118,6 +119,20 @@ func (c ConfigData) GetScanJobTolerations() ([]corev1.Toleration, error) {
 	return scanJobTolerations, err
 }
 
+func (c ConfigData) GetScanJobPodSecurityContext() (*corev1.PodSecurityContext, error) {
+	if c[keyScanJobPodSecurityContext] == "" {
+		return nil, nil
+	}
+
+	scanJobPodSecurityContext := &corev1.PodSecurityContext{}
+	err := json.Unmarshal([]byte(c[keyScanJobPodSecurityContext]), &scanJobPodSecurityContext)
```
erikgb, line 128:
```suggestion
	err := json.Unmarshal([]byte(c[keyScanJobPodSecurityContext]), scanJobPodSecurityContext)
```

nerzhul: fixed :) ty

nerzhul:
Okay i hope we can merge this before you change the whole architecture :) who are maintainers ? in the changelog it seems you are 2 currently active people

josedonizetti:
@nerzhul Thanks for the PR, didn't have time to look/test this yet, but I think your change here is addressing what was asked here? https://github.com/aquasecurity/trivy-operator/issues/121

chen-keinan:
@nerzhul thanks you for contributing , I'm find with keeping the `podSecurityContext` in json format to be consist with this type of data.
I'll approve it later after I'll test it and make sure it works find.

chen-keinan (approved):

chen-keinan:
I assume it will solve the most cases where it is required in pod level however not if it is required for a specific init container

josedonizetti (commented):
Looks good! But I do have a comment/question. Considering trivy will scan and report misconfiguration for pods not having a security context with `runAsUser`,  `runAsGroup` and  `runAsNonRoot`. 

https://avd.aquasec.com/misconfig/kubernetes/general/avd-ksv-0012/
https://avd.aquasec.com/misconfig/kubernetes/general/avd-ksv-0020/
https://avd.aquasec.com/misconfig/kubernetes/general/avd-ksv-0021/

Why don't we have a default value for `podTemplateSecurityContext` considering the AVDs above?

josedonizetti (approved):
I'll create an issue to discuss further if we should have default or not. :)

nerzhul:
@josedonizetti it's a very good comment, i think we should have best security options by default, i just kept compat with existing when writing this pr :)

erikgb:
> @josedonizetti it's a very good comment, i think we should have best security options by default, i just kept compat with existing when writing this pr :)

It will be a breaking change for users on Openshift, but I do not oppose adding a default according to K8s security best practices. In this case, I think Openshift can improve. 😉 But it should be mentioned in the release notes as a breaking change IMO.

Issue #373, opened 2022-07-20
Operator doesn't have permissions to list deployments


erikgb:
**What steps did you take and what happened:**

After upgrading trivy-operator to the latest release (0.1.4), I notice this in our logs:

````
W0720 09:22:13.597998 1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:trivy-system:trivy-operator" cannot list resource "deployments" in API group "apps" at the cluster scope
````

**What did you expect to happen:**

No errors in logs.

**Anything else you would like to add:**

I am pretty sure that this bug was introduced by https://github.com/aquasecurity/trivy-operator/pull/215/files, and I will look into where access to deployments seems required, and either remove the requirement or update the clusterrole.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.1.4
- Kubernetes version (use `kubectl version`): RKE2 1.23.8
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): RHEL 8.5 on K8s hosts

josedonizetti:
@erikgb Not sure as well from the point of view of tighten RBAC, but maybe it does make sense to only scan a rs if there is one active on the cluster? But not a high priority change I would say, perhaps raising an issue to further discuss the idea with the community?

erikgb:
> @erikgb Not sure as well from the point of view of tighten RBAC, but maybe it does make sense to only scan a rs if there is one active on the cluster? Not a high priority change I would say, perhaps raising an issue to further discuss the idea with the community?

@josedonizetti I can create an issue for this, sure. But then I think it should apply for all Pod-controller resources. So not only replicasets, but also statefulsets+++.

PR #376, opened 2022-07-20
refactor: make IsActiveReplicaSet type safe

erikgb:
## Description

Just a small refactoring to make `IsActiveReplicaSet` type safe. Discovered during investigation of https://github.com/aquasecurity/trivy-operator/issues/373.

I also think we should look into avoiding this logic completely, ref. https://github.com/aquasecurity/trivy-operator/issues/237.

## Related issues
- Relates to https://github.com/aquasecurity/trivy-operator/issues/373

## Checklist

### DIFF
--- a/pkg/configauditreport/controller/controller.go
+++ b/pkg/configauditreport/controller/controller.go
@@ -193,8 +193,7 @@ func (r *ResourceController) reconcileResource(resourceKind kube.Kind) reconcile
 		}
 
 		if r.Config.ConfigAuditScannerScanOnlyCurrentRevisions && resourceKind == kube.KindReplicaSet {
-			controller := metav1.GetControllerOf(resource)
-			activeReplicaSet, err := r.IsActiveReplicaSet(ctx, resource, controller)
+			activeReplicaSet, controller, err := r.IsActiveReplicaSet(ctx, resource.(*appsv1.ReplicaSet))
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed checking current revision: %w", err)
 			}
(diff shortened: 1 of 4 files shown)


pkg/configauditreport/controller/controller.go
```diff
@@ -193,8 +193,7 @@ func (r *ResourceController) reconcileResource(resourceKind kube.Kind) reconcile
 		}
 
 		if r.Config.ConfigAuditScannerScanOnlyCurrentRevisions && resourceKind == kube.KindReplicaSet {
-			controller := metav1.GetControllerOf(resource)
-			activeReplicaSet, err := r.IsActiveReplicaSet(ctx, resource, controller)
+			activeReplicaSet, controller, err := r.IsActiveReplicaSet(ctx, resource.(*appsv1.ReplicaSet))
```
chen-keinan, line 196: why it need to be casted  ? and if you still think its  needed then it needed to be check
```suggestion
                        rsResource,ok := resource.(*appsv1.ReplicaSet)
                        if !ok {
                                       return ctrl.Result{}, fmt.Errorf("resource failed casting to replica set: %w", err)
                                    }
			activeReplicaSet, controller, err := r.IsActiveReplicaSet(ctx, rsResource)
```

chen-keinan (changes requested):

pkg/configauditreport/controller/controller.go
erikgb: So the conditional right above is not enough (last part)? How come?

`r.Config.ConfigAuditScannerScanOnlyCurrentRevisions && resourceKind == kube.KindReplicaSet`

erikgb: I find it a bit odd to have a func named `IsActiveReplicaSet` not accepting `ReplicaSet` as param.

chen-keinan: still doesn't answer the question why cast it  if not needed?  also changing the method naming probably will be cheeper

Issue #370, opened 2022-07-20
Suggestion for refactoring: Split WorkloadController


erikgb:
I think the WorkloadController struct has too much responsability, as it is actually two controllers in one struct. This makes the [Go file](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/vulnerabilityreport/controller.go) large, which could be fixed by just splitting funcs between files. But for testing, it would be much easier if we had two structs, one for each controller: `WorkloadController` (create scan jobs from workload) and `ScanJobController` (creating reports from finished scan jobs).

chen-keinan:
its very low priority , I'll put in the backlog in the meantime try to manage with out it

PR #380, opened 2022-07-21
refactor: offload some code from workload_controller.go

erikgb:
## Description

This is a minor refactoring moving some code out of `controller.go`/`workload_controller.go`. As discussed with @chen-keinan this must be done stepwise, and I hope this is an ok first step. I attempted the full refactoring suggested in https://github.com/aquasecurity/trivy-operator/issues/370, and the change in this PR seems like a good way forward.

## Related issues
- Relates to https://github.com/aquasecurity/trivy-operator/issues/370

## Checklist

### DIFF
--- a/pkg/kube/object.go
+++ b/pkg/kube/object.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
+
 	"github.com/aquasecurity/trivy-operator/pkg/trivyoperator"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -175,6 +177,15 @@ func (ci ContainerImages) AsJSON() (string, error) {
 
 func (ci ContainerImages) FromJSON(value string) error {
 	return json.Unmarshal([]byte(value), &ci)
+}
+
+func (ci ContainerImages) DeepEqualTo(m map[string]bool) bool {
+	expected := map[string]bool{}
+	for containerName := range ci {
+		expected[containerName] = true
+	}
+
+	return reflect.DeepEqual(m, expected)
 }
 
 func ObjectRefFromKindAndObjectKey(kind Kind, name client.ObjectKey) ObjectRef {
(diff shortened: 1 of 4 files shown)


pkg/kube/object.go
```diff
@@ -177,6 +179,15 @@ func (ci ContainerImages) FromJSON(value string) error {
 	return json.Unmarshal([]byte(value), &ci)
 }
 
+func (ci ContainerImages) DeepEqualTo(m map[string]bool) bool {
```
erikgb, line 182: The name of this func is not good, and the internals can probably also be improved. But I suggest to move the func here, with `ContainerImages` as a receiver.

josedonizetti (commented):
@erikgb This seems like a good step. Though, considering it is moving unexported methods to be exported, I would prefer we cover them with tests. A quick look on it and I'm not sure we should have the methods `HasVulnerabilityReports` and `HasSecretReports` coupled with `FindOwner`, and probably adding tests will help us understand it. 
WDYT?

erikgb:
> @erikgb This seems like a good step. Though, considering it is moving unexported methods to be exported, I would prefer we cover them with tests. A quick look on it and I'm not sure we should have the methods `HasVulnerabilityReports` and `HasSecretReports` coupled with `FindOwner`, and probably adding tests will help us understand it. WDYT?

I agree with you here. Adding tests will never harm. But I want @chen-keinan to tell me what he wants before I spend/waste time on this - as I am bit tired of suggesting improvements that seem to have no chance of being merged.

chen-keinan:
@LGTM , I agree with @josedonizetti let`s add tests to cover it.

erikgb:
Looking a bit on adding tests, I think we should not add tests in this PR - as this will have to refactored more. And adding unit tests will just be a (double) waste of time when the unit is going to change more. What I really want is good integration tests that can assure we don't break anything during refactoring.

So if you insist on adding tests to something currently untested in this PR, I'll just close the PR. It's really up to you....

erikgb:
It is actually more appealing to rewrite the whole thing regarding "hasReports" by adding indexes and perform "smart" queries, ref. https://github.com/aquasecurity/trivy-operator/issues/381.

PR #388, opened 2022-07-21
fix: ensure that report name is a valid resource name

### LINKED ISSUE #343: Attempt to create ClusterRbacAssessmentReport with invalid name

I just booted up a fresh cluster with trivy-operator and some other stuff installed. It seems like our ClusterRbacAssessmentReport-reconciler is attempting to create a report with an invalid name:

````
{"level":"error","ts":1658084310.4691162,"msg":"Reconciler error","controller":"clusterrole","controllerGroup":"rbac.authorization.k8s.io","controllerKind":"ClusterRole","clusterRole":{"name":"cluster-crd-clusterRole"},"namespace":"","name":"cluster-crd-clusterRole","reconcileID":"a663541b-97ff-4c86-99c2-bdf2fbe7571e","error":"ClusterRbacAssessmentReport.aquasecurity.github.io \"clusterrole-cluster-crd-clusterRole\" is invalid: metadata.name: Invalid value: \"clusterrole-cluster-crd-clusterRole\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234"}
````

Resources scanned without errors.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.1.3
- Kubernetes version (use `kubectl version`): RKE2 1.23.8
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): RHEL 8 on K8s host

### PR DESCRIPTION
**Pull request by erikgb**
## Description

This adds the required extra validation to report names to fix #343. Also refactor the code to avoid duplication in this area.

I know plural package names are discouraged in Golang, but naming things is hard. Please let me know if you have any constructive input to naming.

## Related issues
- Close #343

## Checklist

### DIFF
--- a/pkg/operator/reports/name.go
+++ b/pkg/operator/reports/name.go
@@ -1,0 +1,36 @@
+package reports
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aquasecurity/trivy-operator/pkg/kube"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NameFromController(controller client.Object) string {
+	kind := controller.GetObjectKind().GroupVersionKind().Kind
+	name := controller.GetName()
+	reportName := fmt.Sprintf("%s-%s", strings.ToLower(kind), name)
+	if isValidName(reportName) {
+		return reportName
+	}
+	return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(name))
+}
+
+func NameFromControllerContainer(controller client.Object, container string) string {
+	kind := controller.GetObjectKind().GroupVersionKind().Kind
+	name := controller.GetName()
+	reportName := fmt.Sprintf("%s-%s-%s", strings.ToLower(kind), name, container)
+	if isValidName(reportName) {
+		return reportName
+	}
+
+	return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(name+"-"+container))
+}
+
+func isValidName(name string) bool {
+	// We also use report names as label values, so must validate both valid resource name and label value
+	return len(validation.IsDNS1123Subdomain(name)) == 0 && len(validation.IsValidLabelValue(name)) == 0
+}
--- a/pkg/configauditreport/builder.go
+++ b/pkg/configauditreport/builder.go
@@ -3,15 +3,14 @@ package configauditreport
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
 	"github.com/aquasecurity/trivy-operator/pkg/kube"
+	"github.com/aquasecurity/trivy-operator/pkg/operator/reports"
 	"github.com/aquasecurity/trivy-operator/pkg/trivyoperator"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -52,13 +51,7 @@ func (b *ReportBuilder) Data(data v1alpha1.ConfigAuditReportData) *ReportBuilder
 }
 
 func (b *ReportBuilder) reportName() string {
-	kind := b.controller.GetObjectKind().GroupVersionKind().Kind
-	name := b.controller.GetName()
-	reportName := fmt.Sprintf("%s-%s", strings.ToLower(kind), name)
-	if len(validation.IsValidLabelValue(reportName)) == 0 {
-		return reportName
-	}
-	return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(name))
+	return reports.NameFromController(b.controller)
 }
 
 func (b *ReportBuilder) GetClusterReport() (v1alpha1.ClusterConfigAuditReport, error) {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Review by josedonizetti — APPROVED** [review]
Looks good Erik! :) Can't think of a different name for `reports`, though It looks to me it makes more sense as `pkg/reports` than `pkg/operator/reports` as it is a code common to all reports which are all at the `pkg` level.

**Comment by erikgb** [comment]
> Looks good Erik! :) Can't think of a different name for `reports`, though It looks to me it makes more sense as `pkg/reports` than `pkg/operator/reports` as it is a code common to all reports which are all at the `pkg` level.

Thinking a bit more about it, I think resource names are important enough to actually let the code reside under `pkg/apis`. But it is technically not part of the API. It seems like there is a lot of duplicated code in this area, so I think it makes sense to add some kind of package for sharing code between the different "reports".

**Review by JAORMX — APPROVED** [review]
lol, I've seen this validation in one too many controllers. I wonder why there isn't yet validation code being part of `controller-runtime` or a library of the sort. Anyway, thanks for adding this!

**Comment by erikgb** [comment]
> lol, I've seen this validation in one too many controllers. I wonder why there isn't yet validation code being part of `controller-runtime` or a library of the sort. Anyway, thanks for adding this!

The thing is that this "depends". We've never needed this in our internal controllers. The funcs to validate are there, you just have to use them correctly, depending on your use case.

File: pkg/operator/reports/name.go [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aquasecurity/trivy-operator/pkg/kube"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NameFromController(controller client.Object) string {
```
chen-keinan, line 12: I would change the func name to be more clear to `NameController`

pkg/operator/reports/name.go
```diff
@@ -0,0 +1,36 @@
...
+	kind := controller.GetObjectKind().GroupVersionKind().Kind
+	name := controller.GetName()
+	reportName := fmt.Sprintf("%s-%s", strings.ToLower(kind), name)
+	if isValidName(reportName) {
+		return reportName
+	}
+	return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(name))
+}
+
+func NameFromControllerContainer(controller client.Object, container string) string {
```
chen-keinan, line 22: I would change the func name to be more clear to `ControllerContainerName`

pkg/operator/reports/name.go
```diff
@@ -0,0 +1,36 @@
...
+	kind := controller.GetObjectKind().GroupVersionKind().Kind
+	name := controller.GetName()
+	reportName := fmt.Sprintf("%s-%s-%s", strings.ToLower(kind), name, container)
+	if isValidName(reportName) {
+		return reportName
+	}
+
+	return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(name+"-"+container))
```
chen-keinan, line 30:
```suggestion
        return fmt.Sprintf("%s-%s", strings.ToLower(kind), kube.ComputeHash(fmt.Sprintf("%s-%s", name, container)))

```

chen-keinan (changes requested):

pkg/operator/reports/name.go
erikgb: I don't think we should use fmt here, and leave it as it is. It would be better if the `ComputeHash` func accepted a varadic argument, but that would be a breaking change..

chen-keinan: explain why not?

erikgb: > explain why not use fmt?

1. Because the code was not using `fmt` before this refactoring (and fix)
2. Using `fmt.Sprintf` involves a lot of code for performing advanced formatting that is not needed here. The purpose here is to get something unique from the `name` and `container` values.

chen-keinan: IMO it make sense to use formatting , it make it more clean, clear (even if it wan't like it before) , same as we use it other places in the code rather than connecting with `+`.

pkg/operator/reports/name.go
erikgb: Usually I try to name methods by how they are used, and I think `reports.NameFromController` reads better than `reports.NameController`. Remember that the parameter is just `client.Object`. It would be even better if the package wasn't named with plural (`report`), but that breaks linting.

josedonizetti: Yeah, also prefer `NameFromController`, `NameController` doesn't give the feeling a name will be returned, maybe if we add an action before like `CreateNameFromController`? wdyt?

pkg/operator/reports/name.go
josedonizetti: This name I also didn't like, though I like the consistency with the other method. When I was looking, I actually thought it could have been `NameFromControllerAndContainer`. Though, I would keep it closer to whatever is decided on for the method above.

pkg/operator/reports/name.go
josedonizetti: wdyt about hashing the `reportName` instead of constructing a new one? eg: `kube.ComputeHash(reportName)`
we would be hashing the "real" name that was invalid, but I think the hash wouldn't break because of it.

Issue #400, opened 2022-07-25
ConfigAuditReport error on kubernetes 1.19


josedonizetti:
```
{"level":"error","ts":1658781042.966878,"msg":"Reconciler error","controller":"replicaset","controllerGroup":"apps","controllerKind":"ReplicaSet","replicaSet":{"name":"app-7d8c87ffd5","namespace":"default"},"namespace":"default","name":"app-7d8c87ffd5","reconcileID":"8c4721e4-2e33-47e7-b85f-818def0ecf73","error":"ConfigAuditReport.aquasecurity.github.io \"replicaset-app-7d8c87ffd5\" is invalid: report.updateTimestamp: Invalid value: \"null\": report.updateTimestamp in body must be of type string: \"null\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/josedonizetti/code/trivy-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/josedonizetti/code/trivy-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234"}
{"level":"error","ts":1658781043.5101628,"msg":"Reconciler error","controller":"replicaset","controllerGroup":"apps","controllerKind":"ReplicaSet","replicaSet":{"name":"app-7d8c87ffd5","namespace":"default"},"namespace":"default","name":"app-7d8c87ffd5","reconcileID":"5fc084ef-5609-40e0-b947-ca27eeee1211","error":"ConfigAuditReport.aquasecurity.github.io \"replicaset-app-7d8c87ffd5\" is invalid: report.updateTimestamp: Invalid value: \"null\": report.updateTimestamp in body must be of type string: \"null\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/josedonizetti/code/trivy-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/josedonizetti/code/trivy-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234"}
[...]

PR #401, opened 2022-07-25
fix: set configauditreport updatetimestamp

josedonizetti:

## Description

On k8s 1.19 it is not accepted a zero value metav1.Time, so we need to set it when creating the config audit report like we do for the vulnerability report -> https://github.com/aquasecurity/trivy-operator/blob/main/pkg/plugins/trivy/plugin.go#L1363

## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/400

### DIFF
--- a/pkg/configauditreport/controller/controller.go
+++ b/pkg/configauditreport/controller/controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+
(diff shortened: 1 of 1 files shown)


chen-keinan (approved):

chen-keinan:
LGTM 🚀

PR #417, opened 2022-08-12
Extraction of log reading logic to defer its close

1003n40:
## Description
What we currently do is that we fetch the stream, but when error occurs while unmarshalling we can't close the log reader stream, which stays in memory while gc finds and deletes it. As an idea was to extract the logic to standalone function which does the fetching of log buffer and defers on closing it.

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -42,9 +42,9 @@ operator:
   # vulnerabilityScannerReportTTL the flag to set how long a vulnerability report should exist. "" means that the vulnerabilityScannerReportTTL feature is disabled
   vulnerabilityScannerReportTTL: "24h"
   # configAuditScannerEnabled the flag to enable configuration audit scanner
-  configAuditScannerEnabled: false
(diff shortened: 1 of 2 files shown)


chen-keinan:
@1003n40 thank you for the contribution , did you meant to change the `values.yaml` file

1003n40:
@chen-keinan This isn't related to any code with the `values.yaml`, it is related to possible OOM of the operator when handling multiple logs that are with cut off logs (aka. json outputs).

chen-keinan:
> @chen-keinan This isn't related to any code with the `values.yaml`

OK , Can you please remove the changes from the values.yaml on this PR.

1003n40:
> > @chen-keinan This isn't related to any code with the `values.yaml`
> 
> OK , Can you please remove the changes from the values.yaml on this PR.

Ohh my error :).

chen-keinan:
LGTM 🚀

chen-keinan (approved):

Issue #206, opened 2022-06-28
trivy-operator - error: json: cannot unmarshal number into Go value of type trivy.ScanReport


flngen:
**What steps did you take and what happened:**

Installed trivy-operator 0.0.8 via helm on k8s v1.19.8. Ran:

```
kubectl logs -n trivy-system deployment/trivy-operator 
```
and observed the following in the log:

"stacktrace": "sigs.k8s.io/controller-runtime/pkg/interna/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.1/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/interna/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.1/pkg/internal/controller/controller.go:234"

**What did you expect to happen:**

I expected to see vulnerability reports when running:

```
kubectl get vulnerabilityreports --all-namespaces -o wide
```

**Anything else you would like to add:**

I do see that vulnerability reports (scan jobs) run successfully and report vulns in the logs, however I am unable to retrieve them with the aforementioned kubectl command

**Environment:**

- Trivy-Operator version 0.0.8
- Kubernetes version 1.19.8:

chen-keinan:
@flngen can you provide more info.
- are you getting this output `No resources found` ? 
- are you able to confirm that a vulnerability report has been generated via `kubectl get vulnerabilityreports --all-namespaces` ?

flngen:
I get a "No resources found"; however if I run: kubectl -n trivy-system get pods, I see the "scan" pods listed, and when I run $ kubectl -n trivy-system logs <pod-name>, I see vulnerabilities in json format.

flngen:
Ok I think we're talking in circles here. I get what you're saying that they're supposed to be persisted. That's the whole point, they are not. The error log is what I posted initially on the log. Please reread it. Something about unmarshalling something. 

The pods that start with "scan-<something-random>" for certain contain vulnerabilities in their logs. When you run the command I mention in the previous post, you see the full trivy vuln report, with a bunch of CVEs. However, I'd like to see them when I run ```kubectl get vulnerabilityreports --all-namespaces -o wide```. That's the whole point of this bug report.

chen-keinan:
[...]

PR #431, opened 2022-08-20
fix/compress trivy scan result

### LINKED ISSUE #422: use compression to reduce trivy output report size to avoid hitting container logs max size
It has been found out that trivy scan job produces scanning results bigger than pod max log default value (10MB), as result log rotation happens and operator is unable to unmarshal results from pod logs.

- suggested  workaround : Increase the [kubelet default ](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) `--container-log-max-size`

- long-term solution will be to write the report output to the pod filesystem and stream it 
Related: #206

chen-keinan:
## Related issues
- Close #422 

- Related #206

## Checklist
- [x] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant


pkg/plugins/trivy/plugin.go
```diff
@@ -661,17 +664,10 @@ func (p *plugin) getPodSpecForStandaloneMode(ctx trivyoperator.PluginContext, co
...
 			Args: []string{
-				"--cache-dir",
-				"/tmp/trivy/.cache",
-				"--quiet",
-				"image",
-				"--skip-update",
-				"--format",
-				"json",
-				optionalMirroredImage,
+				"-c", fmt.Sprintf("trivy image %s --cache-dir /tmp/trivy/.cache --quiet  --skip-update --format json > /tmp/scan/result.json &&  bzip2 -c /tmp/scan/result.json | base64", optionalMirroredImage),
```
JAORMX, line 670: Has the value of `optionalMirroredImage` been validated beforehand? Just making sure someone doesn't do anything funky with that now that we're using it for string interpolation.

pkg/plugins/trivy/plugin.go
```diff
@@ -1026,16 +1019,10 @@ func (p *plugin) getPodSpecForClientServerMode(ctx trivyoperator.PluginContext,
...
 			},
 			Args: []string{
-				"--quiet",
-				"image",
-				"--format",
-				"json",
-				"--server",
-				trivyServerURL,
-				optionalMirroredImage,
+				"-c", fmt.Sprintf("trivy image %s --quiet --format json --server %s > /tmp/scan/result.json &&  bzip2 -c /tmp/scan/result.json | base64", optionalMirroredImage, trivyServerURL),
```
JAORMX, line 1025: have the `optionalMirroredImage` and `trivyServerURL` been validated beforehand? Someone could insert arbitrary commands through those.

JAORMX (changes requested):
this LGTM, I just have a couple of minor questions that I left in-line, and that's it.

pkg/plugins/trivy/plugin.go
chen-keinan: @JAORMX thanks for the input , looking at it.

JAORMX (approved):
/lgtm

pkg/utils/compressutil.go
```diff
@@ -0,0 +1,30 @@
...
+	_, err := io.Copy(uncompressedWriter, bz2Reader)
+	if err != nil {
+		return nil, err
+	}
+	return uncompressedWriter, nil
+}
+
+// Base64Decode accept encoded reader and base64 decode it
+func Base64Decode(EncodedReader io.Reader) ([]byte, error) {
+	EncodedBytes, err := ioutil.ReadAll(EncodedReader)
```
josedonizetti, line 25:
```suggestion
	encodedBytes, err := ioutil.ReadAll(EncodedReader)
```

josedonizetti (approved):
only found a typo

JAORMX (approved):

chen-keinan:
holding with this PR as we hit the `etcd` limit `etcdserver: request is too large` issue

pkg/utils/compressutil.go
```diff
@@ -0,0 +1,30 @@
...
+	uncompressedWriter := new(bytes.Buffer)
+	_, err := io.Copy(uncompressedWriter, bz2Reader)
+	if err != nil {
+		return nil, err
+	}
+	return uncompressedWriter, nil
+}
+
+// Base64Decode accept encoded reader and base64 decode it
+func Base64Decode(EncodedReader io.Reader) ([]byte, error) {
```
josedonizetti, line 24: typo

```suggestion
func Base64Decode(encodedReader io.Reader) ([]byte, error) {
```

JAORMX:
LGTM

shipit! 😄

PR #463, opened 2022-08-26
feat: add securityContext for initContainers

FelixKampfer:
## Description
This change sets a default securityContext for initContainers. This is useful because some admission controllers reject containers with no securityContext. 
An alternative implementation could see `scanJob.podTemplateSecurityContext` being passed to pkg/plugins/trivy/plugin.go's `GetScanJobSpec`, but I thought this might be simpler.

## Related issues
- Improves #121 

## Checklist


chen-keinan:
@FelixKampfer thank you for the contribution, I wonder if it is worth making it configurable with default values set (as proposed). Allowing users the ability to overwrite values , wdyt ?

FelixKampfer:
What about [the hardcoded securityContext of the scan containers](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/plugins/trivy/plugin.go#L679)? What would happen to the current hardcoded value?

Could I make these configurable with the same property key? 
E.g.   `scanJobPodTemplateContainerSecurityContext`, so that the existing `scanJobPodTemplateSecurityContext` doesn't need to be modified.

chen-keinan:
@FelixKampfer do we need this PR after merging #477 ?

FelixKampfer:
No we do not, thanks!

PR #473, opened 2022-08-29
refactor: reuse controllers code

chen-keinan:

## Description
reuse controllers code

### DIFF
--- a/pkg/operator/workload/helper.go
+++ b/pkg/operator/workload/helper.go
@@ -12,18 +12,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func Inspect(ctx context.Context, resource client.Object, or kube.ObjectResolver, scanOnlyCurrentRevisions bool, log logr.Logger) (bool, error) {
+func SkipProcessing(ctx context.Context, resource client.Object, or kube.ObjectResolver, scanOnlyCurrentRevisions bool, log logr.Logger) (bool, error) {
 	controller := metav1.GetControllerOf(resource)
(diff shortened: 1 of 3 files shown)


pkg/operator/workload/helper.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/go-logr/logr"
+	"golang.org/x/net/context"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Inspect(ctx context.Context, resource client.Object, or kube.ObjectResolver, scanOnlyCurrentRevisions bool, log logr.Logger) (bool, error) {
```
josedonizetti, line 15: Can we improve the name? Inspect doesn't give any idea about what this func is doing?

chen-keinan: @josedonizetti sure , the other thing on my mind was `validate` happy for suggestion

josedonizetti: @chen-keinan I think what we are doing here is `skipping`, which I also prefer because would change the check to `if skipping` instead of `if !skipping`. So I'm thinking:

```
skip, err = workload.SkipProcessing() // or ShouldSkip()
if skip {
return
}
```

wdyt?

josedonizetti (approved):
Good to go once CI is green :)

Issue #70, opened 2022-05-31
integration with postee


chen-keinan:
Add support for integration with [postee](https://github.com/aquasecurity/postee) :
- support configuration setting for postee
- add capability to wire reports via postee
- add capability to filter reports data passed to postee

simar7:
hi @chen-keinan - is this currently something that's being worked upon? this would be very handy to integrate trivy and postee in a k8s environment. I can send a PR if needed.

chen-keinan:
@simar7 go ahead  🚀  I'll love to see it going on

simar7:
hi @chen-keinan - just wanted to solicit some feedback regarding how we can go about implementing this before I start writing code and also to validate my understanding regarding the Trivy-operator codebase.

1. Currently I see there are 2 kinds of reports being generated: `clustercompliancereport` and `configauditreport`. They both seem to live as CRDs. Which one would make the most sense to send over to Postee? Both?
2. Is there anything sensitive stored as part of the results in the CRD? 
3. How often should these reports be sent over? My initial idea is to wire the sending over of reports every time it is written into the CRD as done here: https://github.com/aquasecurity/trivy-operator/blob/38d04cdd62d8c6843eabf264207b8ead2a7e735a/pkg/configauditreport/io.go#L66



Open to ideas.

chen-keinan:
> hi @chen-keinan - just wanted to solicit some feedback regarding how we can go about implementing this before I start writing code and also to validate my understanding regarding the Trivy-operator codebase.
> 
> 1. Currently I see there are 2 kinds of reports being generated: `clustercompliancereport` and `configauditreport`. They both seem to live as CRDs. Which one would make the most sense to send over to Postee? Both?
> 2. Is there anything sensitive stored as part of the results in the CRD?
> 3. How often should these reports be sent over? My initial idea is to wire the sending over of reports every time it is written into the CRD as done here: https://github.com/aquasecurity/trivy-operator/blob/38d04cdd62d8c6843eabf264207b8ead2a7e735a/pkg/configauditreport/io.go#L66
> 
> Open to ideas.

1. We have today 8 types of CRDs the last two are not active today (will be back soon)
- `clusterconfigauditreports`
- `clusterrbacassessmentreports`
-  `configauditreports`
- `exposedsecretreports`
- `rbacassessmentreports`
 - `vulnerabilityreports`
 - `clustercompliancedetailreports`
 - `clustercompliancereports`
 
[...]

PR #476, opened 2022-09-01
feat(op): Add support for sending reports to a webhook

simar7:
## Description

This PR adds the ability to send reports to a webhook endpoint.

## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/70


pkg/operator/etc/config.go
```diff
@@ -34,6 +34,8 @@ type Config struct {
 	LeaderElectionEnabled                        bool           `env:"OPERATOR_LEADER_ELECTION_ENABLED" envDefault:"false"`
 	LeaderElectionID                             string         `env:"OPERATOR_LEADER_ELECTION_ID" envDefault:"trivyoperator-lock"`
 	ExposedSecretScannerEnabled                  bool           `env:"OPERATOR_EXPOSED_SECRET_SCANNER_ENABLED" envDefault:"true"`
+	WebhookBroadcastURL                          string         `env:"OPERATOR_WEBHOOK_BROADCAST_URL"`
+	WebhookBroadcastTimeout                      *time.Duration `env:"OPERATOR_WEBHOOK_BROADCAST_TIMEOUT" envDefault:"30s"`
```
simar7, line 38: At the moment I wanted to keep it simple and make only the `URL` and the `timeout` configurable by the user. This would imply that we would send all reports at all times as they are generated/updated.

chen-keinan:
@simar7 maybe it is possible to create a new `controller` for `webhook` only which will reconcile reports (vulnerability and etc) when created and send webhook api call , wdyt ?

simar7:
> @simar7 thank you for working on it ,maybe it is possible to create a new `controller` for `webhook` only which will reconcile reports (vulnerability and etc) when created and send webhook api call , wdyt ?

So I tried to implement your suggestion but quickly ran into a situation where in the case of a `vulnerabilityreport` we'll have to essentially duplicate the logic that exists today for `reconcileJob` and `processCompleteScanJob`. 

Another thought I had was how the scheduling would take place between controllers that are "producers" like `vulnerabilityreport` and controllers that are "consumers" like the new `webhook` controller. My understanding is that they are event based so it is likely that the webhook consumer controller can consume an old report and send it out while the producer has not finished / or has updated the report in the meantime.

The easiest way I see is to keep it simple and consume when the report is produced. We could certainly move the webhook specific logic to its own package under the `/pkg` directory. 

Another way to reuse this new package would be to use generics as I have in this PR. Essentially we just need the input (in this case a "report") to be marshalable. So the following prototype would work for all kinds of reports, including (but not limited to): `vulnerabilityreport`, `configauditreport` etc.

```
func sendReports[T any](reports []T, endpoint string, timeout time.Duration) error 
```

Let me know what you think or maybe I have missed something.

pkg/webhook/webhookreporter.go
```diff
@@ -0,0 +1,88 @@
...
+	reports := []client.Object{
+		&v1alpha1.VulnerabilityReport{},
+		&v1alpha1.ExposedSecretReport{},
+	}
+
+	for _, reportType := range reports {
+		err = ctrl.NewControllerManagedBy(mgr).
+			For(reportType, builder.WithPredicates(
+				predicate.Not(predicate.IsBeingTerminated),
+				installModePredicate)). // TODO: Is there a predicate that determines existance?
```
simar7, line 44: @chen-keinan - is there a predicate that can determine existence? For instance, if I add to the list of supported reports slice, `clustercompliancereport` and let's say even if the report generation is enabled as part of the trivy operator config but hasn't been generated just yet, this is what I see in the logs:

```

1.6625969315044768e+09  ERROR   controller-runtime.source       if kind is a CRD, it should be installed before calling Start   {"kind": "ClusterComplianceDetailReport.aquasecurity.github.io", "error": "no matches for kind \"ClusterComplianceDetailReport\" in version \"aquasecurity.github.io/v1alpha1\""}
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1.1
        /Users/simar/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/source.go:139
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext
        /Users/simar/go/pkg/mod/k8s.io/apimachinery@v0.24.4/pkg/util/wait/wait.go:233
k8s.io/apimachinery/pkg/util/wait.WaitForWithContext
        /Users/simar/go/pkg/mod/k8s.io/apimachinery@v0.24.4/pkg/util/wait/wait.go:660
k8s.io/apimachinery/pkg/util/wait.poll
        /Users/simar/go/pkg/mod/k8s.io/apimachinery@v0.24.4/pkg/util/wait/wait.go:594
k8s.io/apimachinery/pkg/util/wait.PollImmediateUntilWithContext
        /Users/simar/go/pkg/mod/k8s.io/apimachinery@v0.24.4/pkg/util/wait/wait.go:545
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1
        /Users/simar/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/source.go:132
```

chen-keinan: yes , currently the `compliancereports` are disable , ignore these reports for now , we will add it back later

pkg/webhook/webhookreporter.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("failed to marshal reports: %w", err)
+	} else {
+		hc := http.Client{
+			Timeout: timeout,
+		}
+		_, err = hc.Post(endpoint, "application/json", bytes.NewBuffer(b))
+		if err != nil {
+			return fmt.Errorf("failed to send reports to endpoint: %w", err)
+		}
+	}
```
chen-keinan, line 87:
```suggestion
	} 
        hc := http.Client{
	        Timeout: timeout,
        }
        _, err = hc.Post(endpoint, "application/json", bytes.NewBuffer(b))
        if err != nil {
	        return fmt.Errorf("failed to send reports to endpoint: %w", err)
        }
	
```

pkg/webhook/webhookreporter.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				log.V(1).Info("Ignoring cached report that must have been deleted")
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("getting report from cache: %w", err)
+		}
+
+		if err := sendReport(reportType, r.WebhookBroadcastURL, *r.WebhookBroadcastTimeout); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to send vulnerability report: %w", err)
```
chen-keinan, line 69:
```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to send report: %w", err)
```

pkg/vulnerabilityreport/controller.go
```diff
@@ -4,10 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/aquasecurity/trivy-operator/pkg/operator/workload"
```
chen-keinan, line 7: can we revert this file change?

chen-keinan (changes requested):

pkg/vulnerabilityreport/controller.go
simar7: Sure I can and did, although running a standard go fmt linter over this file will re-arrange them as per alphabetical order again as it isn't so today.

pkg/webhook/webhookreporter.go
simar7: Thanks.

chen-keinan:
@simar7 Looks very good , thank you very much for this contribution , one comment , I would add documentation for `webhook` under `integration` section and explain a little bit on usage and of course promote postee

simar7:
> for

Thanks added with this https://github.com/aquasecurity/trivy-operator/pull/476/commits/f90d411a595bff3cff58468770d44eaf99ff2b88?short_path=c6c10fb#diff-c6c10fb184e573db21bc09b5c881e3c6830ccbc75ff763a61276a5ec2225ab89

chen-keinan (approved):

chen-keinan:
LGTM 🚀

PR #523, opened 2022-09-16
feat: add ability to specify which Kubernetes workload resources to scan

### LINKED ISSUE #265: Possibility to Exclude Jobs/cronjobs from scanning
Since jobs/cronjobs represent transient workloads it would be possible to skip/exclude them as it is already possible with complete namespaces. So any configuration option would help here to stop frequent scans in case the frequency for a job is set to a minimum.

@erikgb, @chen-keinan: cc

tks98:
## Description

This PR adds the "targetWorkloads" config field which lets users specify the Kubernetes workload resources they want to scan using the vulnerability-report and config-audit report scanners. By default, all workload resource types are scanned. 

PR adds logic to do this, adds a test case for the config, updates the usage documentation, updates the helm chart and static manifests. 

I had a use-case similar to the related issue and made these modifications for personal use. I saw the related issue today and wanted to open this PR in case it could also benefit the wider community. 

Please let me know if anything else is required, or needs to be changed. Thanks! 

## Related issues
- [Close #265](https://github.com/aquasecurity/trivy-operator/issues/265)

## Checklist


pkg/configauditreport/controller/controller.go
```diff
@@ -105,6 +101,29 @@ func (r *ResourceController) SetupWithManager(mgr ctrl.Manager) error {
...
+			resources = append(resources, resource{kind: kube.KindDaemonSet, forObject: &appsv1.DaemonSet{}, ownsObject: &v1alpha1.ConfigAuditReport{}})
+		case "cronjob":
+			resources = append(resources, resource{kind: kube.KindCronJob, forObject: r.ObjectResolver.GetSupportedObjectByKind(kube.KindCronJob), ownsObject: &v1alpha1.ConfigAuditReport{}})
+		case "job":
+			resources = append(resources, resource{kind: kube.KindJob, forObject: &batchv1.Job{}, ownsObject: &v1alpha1.ConfigAuditReport{}})
+		default:
+			return fmt.Errorf("workload %s is not supported", tw)
+		}
+	}
+
```
chen-keinan, line 126: could we extract it to a method(pkg-->kube-->object) so we could reuse it for all controllers, wdyt ?

deploy/helm/values.yaml
```diff
@@ -14,6 +14,11 @@ targetNamespaces: ""
 # mode, i.e. when the targetNamespaces values is a blank string.
 excludeNamespaces: "kube-system,{{ .Release.Namespace }}"
 
+# targetWorkloads is a comma seperated list of Kubernetes workload resources
+# to be included in the vulnerability and config-audit scans
+# if left blank, all workload resources will be scanned
+targetWorkloads: "Pod,ReplicaSet,ReplicationController,StatefulSet,DaemonSet,CronJob,Job"
```
chen-keinan, line 20: Term thing, make it lower case if we refer it as `resources` (in comments) or keep it like this 
and refer it as `Kinds` 
```suggestion
targetWorkloads: "pod,replicaset,replicationcontroller,statefulset,daemonset,cronjob,job"
```

chen-keinan (changes requested):

chen-keinan:
@tks98 Look's good , thank you for the contribution , I have added couple of comments

pkg/configauditreport/controller/controller.go
tks98: Yes, good idea. I have abstracted that into its own [method](https://github.com/aquasecurity/trivy-operator/pull/523/files#diff-7820ee8c3e9b7e90a347b7fe2e1f81faa3933252bb22d1a272e890490f215072). Let me know if that looks good.

deploy/helm/values.yaml
tks98: Changed these to lower-case in the helm values and updated the static manifest to reflect it.

pkg/kube/object.go
```diff
@@ -662,6 +663,40 @@ func (o *ObjectResolver) GetActivePodsMatchingLabels(ctx context.Context, namesp
...
+	case "replicaset":
+		*r = Resource{Kind: KindReplicaSet, ForObject: &appsv1.ReplicaSet{}, OwnsObject: object}
+	case "replicationcontroller":
+		*r = Resource{Kind: KindReplicationController, ForObject: &corev1.ReplicationController{}, OwnsObject: object}
+	case "statefulset":
+		*r = Resource{Kind: KindStatefulSet, ForObject: &appsv1.StatefulSet{}, OwnsObject: object}
+	case "daemonset":
+		*r = Resource{Kind: KindDaemonSet, ForObject: &appsv1.DaemonSet{}, OwnsObject: object}
+	case "cronjob":
+		*r = Resource{Kind: KindCronJob, ForObject: &batchv1.CronJob{}, OwnsObject: object}
```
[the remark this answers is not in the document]
tks98: This has been updated, thanks!

chen-keinan (approved):

PR #525, opened 2022-09-16
feat: optional namespace

### LINKED ISSUE #487: Specify namespace in Helm Chart Values.yaml manifest different to {{ .Release.Namespace }}
When installing Trivy as a Helm Chart dependency, I would like to set the `trivy-system` namespace in the Values.Yaml manifest.

Currently, the namespace is defined with `{{ .Release.Namespace }}` however, that will take the same namespace as the application/the main Chart in the Helm Chart, which reduces the flexibility to use the trivy-operator with other applications/processes.

Here are the manifests used to install the Chart dependency:
Dependency in Chart.Yaml
```
dependencies:
  - name: trivy-operator
    version: "0.1.9"
    repository: "https://aquasecurity.github.io/helm-charts/"
```

Values for the Trivy-operator Helm Chart, to which I would like to add the namespace
```
trivy-operator:
  trivy:
    ignoreUnfixed: true
```

dirien:
Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

## Description

Add functionality to create a namespace independent from the re…lease namespace. This useful, when using Trivy as dependency in another Helm chart

## Related issues
- Close #487 

## Checklist

### DIFF
--- a/deploy/helm/templates/ns.yaml
+++ b/deploy/helm/templates/ns.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.operator.namespace }}
+{{- if not (lookup "v1" "Namespace" "" .Values.operator.namespace) }}
+{{- if not (eq .Release.Namespace .Values.operator.namespace) }}
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Values.operator.namespace }}
+  labels:
+    {{- include "trivy-operator.labels" . | nindent 4 }}
+
+{{- end }}
+{{- end }}
+{{- end }}
--- a/docs/operator/installation/helm.md
+++ b/docs/operator/installation/helm.md
@@ -54,6 +54,12 @@ except `kube-system` and `trivy-system`:
    kubectl logs deployment/trivy-operator -n trivy-system
    ```
 
+## Install as Helm dependency
+
+There are cases, when potential chart developers want to add the operator as dependency. An example would be the creation of an umbrella chart for an application, which depends on 3d-party charts.
+
+In this case, It maybe not suitable to install the Trivy Operator in the same namespace as the main application. Instead, we can use the Helm value `operator.namespace` to define a namespace where only the operator will be installed. The Operator chart will then either create a new namespace if not existing or use the existing one.
+
 ## Uninstall
 
 You can uninstall the operator with the following command:
--- a/deploy/helm/templates/NOTES.txt
+++ b/deploy/helm/templates/NOTES.txt
@@ -1,4 +1,4 @@
-You have installed Trivy Operator in the {{ .Release.Namespace }} namespace.
+You have installed Trivy Operator in the {{ include "trivy-operator.namespace" . }} namespace.
 It is configured to discover Kubernetes workloads and resources in
 {{ tpl .Values.targetNamespaces . | default "all" }} namespace(s).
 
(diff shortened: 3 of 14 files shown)


chen-keinan:
@dirien Looks good! one comment, I think we need also to update the following values :
- [deployment.yaml L37](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/templates/deployment.yaml#L37)
- [leader_election L41](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/templates/leader_election.yaml#L41)

dirien:
HI @chen-keinan, updated the files. 

Do you think we need to write some documentation for it, apart from the comments in the values.yaml?

chen-keinan:
> HI @chen-keinan, updated the files.
> 
> Do you think we need to write some documentation for it, apart from the comments in the values.yaml?

Good point , I think it will be good to mention it in the [Install docs](https://github.com/aquasecurity/trivy-operator/blob/main/docs/operator/installation/helm.md)

dirien:
Ok will do!

chen-keinan:
@dirien did some more testing:  
- Update the copy template command in `hack/update-static-yaml.sh` to avoid creation of namespace twice 
```
cat $CRD_DIR/*  $HELM_TMPDIR/trivy-operator/templates/* > $STATIC_DIR/trivy-operator.yaml

## if ns.yaml do not exist, cat namespace.yaml to trivy-operator.yaml (avoid duplicate namespace definition)
[ ! -f $HELM_TMPDIR/trivy-operator/templates/ns.yaml ] && cat $STATIC_DIR/namespace.yaml >> $STATIC_DIR/trivy-operator.yaml
``` 

- Add `trivy-operator` labels to ns.yaml

chen-keinan (changes requested):

deploy/helm/templates/ns.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.operator.namespace }}
+{{- if not (lookup "v1" "Namespace" "" .Values.operator.namespace) }}
+{{- if not (eq .Release.Namespace .Values.operator.namespace) }}
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Values.operator.namespace }}
+
```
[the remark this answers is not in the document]
dirien: Done

dirien:
Hi @chen-keinan, 

thanks for the feedback! I added them to the PR!

I wrote also a chapter in the docs for the new feature and would be happy to read yours and @AnaisUrlichs feedback!

chen-keinan:
LGTM 🚀  thank you for these essential contribution.

Holding merge for @AnaisUrlichs to provide feedback on updated docs.

chen-keinan (approved):

docs/operator/installation/helm.md
```diff
@@ -54,6 +54,15 @@ except `kube-system` and `trivy-system`:
...
    ```
 
+## Install as Helm dependency
+
+There are cases, when potential chart developers want to add the operator as dependency. An example would be the creation 
+of an umbrella chart for an application, which depends on 3d-party charts.
+
+In this case, It maybe not suitable to install the Trivy Operator in the same namespace as the main application. Instead, we can use the Helm value 
+`operator.namespace` to define a namespace where only the operator will be installed. The Operator chart will then either create a new namespace
+if not existing or use the existing one.
```
AnaisUrlichs, line 64: there is a weird tap to the next line 😆 otherwise, this is great! Thank you @dirien 🥳

dirien: Done, removed the newlines. Thanks @AnaisUrlichs for reviewing! 🙇

dirien:
Hi @chen-keinan,

last feedback is in and I fixed the merge conflict. Ready when you are.

PR #544, opened 2022-09-23
test: remove unneeded dependency to dyqual

erikgb:
## Description

In https://github.com/aquasecurity/trivy-operator/pull/337, I introduced a dependency to https://github.com/tenstad/dyqual to get better feedback in tests when asserting large struct equality.

We were also using dyqual for the exact same reason, but recently I became aware of the https://github.com/google/go-cmp matchers in Gomega, which provide similar features. This PR eliminates the dependency to dyqual, and when it comes to dependencies, I think we all can agree that "less is more". 😉 

## Related issues
- Relates to #335

## Checklist


JAORMX (approved):

erikgb:
@chen-keinan Any reason this PR seems to be ignored? Anything missing?

JAORMX (approved):

PR #557, opened 2022-10-02
Added ability to use secrets to scan private repo

### LINKED ISSUE #556: Use of ns secrets to authenticate in private repos
Hello!

At the moment `Trivy` using `imagePullSecrets` to authenticate in private repos for image scanning. In my environment there are no imagePullSecrets, because credentials are already placed in my container runtime. That's why `Trivy` cannot scan images from my private repo.

I thought that it would be great to add one more parameter, something like `privateRegistryScanSecretsNames`. This parameter will hold data as map with `namespace:secretname` which can be used to authenticate in private repos.

Also I already have a PoC how it can be implemented. What do you think about it? If it's ok, i will make PR for code review. Thanks!

j1nka:
## Description
Added the ability to use secrets form namespaces rather than from imagePullSecrets to authenticate in private repositories. For example if there is no such secrets (imagePull) and credentials are "stored" in container runtime configuration.

New field - `privateRegistryScanSecretsNames` can be used to set namespaces and secret names. For example: `privateRegistryScanSecretsNames: {"default":"mysecret"}`

## Related issues
- Close #556 

## Checklist
- [X] I've added tests that prove my fix is effective or that my feature works. Tested it locally in my environment

### DIFF
--- a/pkg/kube/secrets.go
+++ b/pkg/kube/secrets.go
@@ -178,6 +178,22 @@ func (r *secretsReader) ListImagePullSecretsByPodSpec(ctx context.Context, spec 
 	return secrets, nil
 }
 
+func (r *secretsReader) GetSecretsFromEnv(ctx context.Context, secretsInfo map[string]string) ([]corev1.Secret, error) {
+	secretsFromEnv := make([]corev1.Secret, 0)
+
+	for ns, secretName := range secretsInfo {
+		var secretFromEnv corev1.Secret
+		err := r.client.Get(ctx, client.ObjectKey{Name: secretName, Namespace: ns}, &secretFromEnv)
+
+		if err != nil && !k8sapierror.IsNotFound(err) {
+			return nil, fmt.Errorf("getting secret by name: %s/%s: %w", ns, secretName, err)
+		}
+
+		secretsFromEnv = append(secretsFromEnv, secretFromEnv)
+	}
+	return secretsFromEnv, nil
+}
+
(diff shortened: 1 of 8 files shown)


pkg/kube/secrets.go
```diff
@@ -187,5 +187,16 @@ func (r *secretsReader) CredentialsByWorkload(ctx context.Context, workload clie
...
+	for ns, secretName := range secretsInfo {
+		var envSecret corev1.Secret
+		err = r.client.Get(ctx, client.ObjectKey{Name: secretName, Namespace: ns}, &envSecret)
+
+		if err != nil && !k8sapierror.IsNotFound(err) {
+			return nil, fmt.Errorf("getting secret by name: %s/%s: %w", ns, secretName, err)
+		}
+
+		imagePullSecrets = append(imagePullSecrets, envSecret)
+	}
```
chen-keinan, line 200: nit: maybe extract it to method

chen-keinan (commented):
It will be also great to update the [private registries tutorial](https://github.com/aquasecurity/trivy-operator/blob/main/docs/tutorials/private-registries.md)

pkg/kube/secrets.go
j1nka: Chen, hi! Added separate method - `GetSecretsFromEnv`

j1nka:
> It will be also great to update the [private registries tutorial](https://github.com/aquasecurity/trivy-operator/blob/main/docs/tutorials/private-registries.md)

Done! Added the `Fourth Option: Define Secrets through Trivy-Operator configuration` in `Tutorials`

PR #563, opened 2022-10-05
Add metrics about CVE vulnerability ID

### LINKED ISSUE #562: Add metrics about CVE vulnerability ID
Currently operator exposed metrics about VulnerabilityReport.
That's a good thing, but there is nothing about CVE. 

I propose to add metrics `trivy_vulnerability_id` which is a sensibly the same as current `trivy_image_vulnerabilities` but with the label `vuln_id` which permit to build dashboard focused en CVE, and made simple prometheus queries to search specific CVE in whole cluster.

In the past we have build something like that based on starboard with a json exporter, but this was hard to maintain and scale. Integrate this feature in operator metrics seems for me to be a good idea.

MatthieuFin:
## Description
This is a proposition about feature request [562](https://github.com/aquasecurity/trivy-operator/issues/562).

Which permit to expose new metrics (enable by default but disabled via operator env variable) about CVE vulnerabilities find in VulnerabilityReports

## Related issues
- Close #562

## Checklist

### DIFF
--- a/deploy/helm/templates/deployment.yaml
+++ b/deploy/helm/templates/deployment.yaml
@@ -12,9 +12,6 @@ spec:
   selector:
     matchLabels:
       {{- include "trivy-operator.selectorLabels" . | nindent 6 }}
-      {{- if .Values.operator.podLabels }}
-      {{- toYaml .Values.operator.podLabels | nindent 6 }}
-      {{- end }}
   template:
     metadata:
       {{- with .Values.podAnnotations }}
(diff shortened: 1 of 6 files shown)


chen-keinan:
@MatthieuFin this change could be a nice addition to `trivy-operator` metrics capability, however I agree with @JAORMX about the impact of metrics cardinality , I would suggest to make it `metricsVulnIdEnabled` disabled by default until we gather enough feedback on it usage.

MatthieuFin:
Hi, yes i'm agree with that too !
ok to set it disabled by default, it will avoid bad surprise on upgrade.

But i have no really good workaround on this issue.
We could perhaps record only CVE with a minimum severity. Or reduce number of label maybe set it as configurable. By example we could only keep namespaces and not image repo / name / version.

chen-keinan:
LGTM 🚀  thank you for the contribution

chen-keinan (approved):

cebidhem:
I was looking for this, but reading the legit concerns I'd like to know if any of you had feedbacks on whether or not cardinality proved to be an issue with this enabled ?

PR #576, opened 2022-10-11
feat: increase the priority of imagePullSecret from the pod spec

### LINKED ISSUE #575: Pod imagePullSecrets priority
Within our organization, we use quay to store docker images of services and to proxy requests to hub.docker.com.

A secret with user credentials is replicated to each namespace. This user has the right to send proxied requests.
This secret is specified as `imagePullSecret` in the `default` service account

In the same namespace, a Pod can be deployed, the image of which is placed in our private quay registry. In this case, we will have two secrets with the same docker registry domain.

When trivy operator [receiving](https://github.com/aquasecurity/trivy-operator/blob/5b9498d9d9fe186fb4861e69b161c9af87d61b38/pkg/kube/secrets.go#L167) `imagePullSecrets` it first gets them from the Pod specification, and then adds the values from the Service Account to the slice

Thus, if the values of the domains in the secrets are the same, then in the final structure, the credentials from the secret specified in the Service Account will "overwrite" the values from the secret specified in the Pod specification.

In our case, as a result, we get an error accessing docker registry when the trivy operator starts scanning.

I looked at how Kubernetes works with imagePullSecrets. [Kubelet](https://github.com/kubernetes/kubernetes/blob/98233be715f78c22c24caf1158e51c4beb1f8644/pkg/kubelet/kubelet.go#L1733-L1739) always uses the value from the [Pod specification](https://github.com/kubernetes/kubernetes/blob/98233be715f78c22c24caf1158e51c4beb1f8644/pkg/kubelet/kubelet_pods.go#L885-L904). In the process of Pod creating `ServiceAccount Admission Controller` [writes](https://github.com/kubernetes/kubernetes/blob/3992eda8e61725c470fb6141a7fe4e7f9ee31ea5/plugin/pkg/admission/serviceaccount/admission.go#L164-L169) the name `imagePullSecret` from ServiceAccount to the Pod specification, if this value was not specified in the Pod manifest.

I think it is necessary to swap imagePullSecrets when filling slice in the source code of the Trivial operator and put the values from the Pod specification at the end. In this case, the behavior will coincide with the logic of kubelet for the specified case.

I will soon create a PR with changes.

verdel:
## Description
Then the values of the domains for docker registry in the pod imagePullSecret and service account imagePullSecret are the same, then in the final structure, the credentials from the secret specified in the Service Account will "overwrite" the values from the secret specified in the Pod specification.

I increase priority of the imagePullSecret from Pod priority. This coincides with the principle of how kubelet working with imagePullSecret.

## Related issues
- Close #575

### DIFF
--- a/pkg/kube/secrets.go
+++ b/pkg/kube/secrets.go
@@ -164,7 +164,7 @@ func (r *secretsReader) ListImagePullSecretsByPodSpec(ctx context.Context, spec 
 	if err != nil && !k8sapierror.IsNotFound(err) {
 		return nil, err
 	}
-	imagePullSecrets = append(imagePullSecrets, sa.ImagePullSecrets...)
+	imagePullSecrets = append(sa.ImagePullSecrets, imagePullSecrets...)
 
 	secrets, err := r.ListByLocalObjectReferences(ctx, imagePullSecrets, ns)
 	if err != nil {
--- a/pkg/kube/secrets_test.go
+++ b/pkg/kube/secrets_test.go
@@ -97,6 +97,38 @@ func TestMapDockerRegistryServersToAuths(t *testing.T) {
 			}),
 		}))
 	})
+
+	t.Run("Test with service account and secret with same registry domain should map container images to Docker authentication credentials from Pod secret", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+
+		var saSecret corev1.Secret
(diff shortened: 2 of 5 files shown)


chen-keinan:
@verdel LGTM 🚀 could you please sign the CLA

chen-keinan (approved):

PR #611, opened 2022-10-24
Add resource labels to metrics

### LINKED ISSUE #478: Ability to read labels from pod/deployements being scanned and add to prometheus metrics
In our clusters we have specific labels that we'd like to add to the metrics generated by `trivy-operator`.

Namely we'd like the ability to assign owner label to these metrics since trying to join these metrics with other metrics to get the owner is non-trivial.

IE, trivy scans doom pod/deployment, then reads owner label. Adds that to the metrics generated.

Let me know if you need more clarification here.

hlastras:
## Description
Add, by configuration, scanned resources labels into the prometheus metrics with a configurable prefix.

## Related issues
- Close #478

## Checklist


hlastras:
The biggest changes have been in the `collector.go` file.

I had to do a little refactor to make it workable. I have moved all the prometheus descriptors, labels and severities that were in global variables, to a struct that will go inside the `ResourcesMetricsCollector`.
This is because the labels are no longer fixed, but instead vary according to the configuration, so they must be generated once we have the configuration data.
I have added a `constructor` for `ResourcesMetricsCollector` to build the labels and the descriptors based on the configuration.

I don't know if I have added enough tests, or if I should add any integration tests.
I haven't added any documentation yet until I'm sure the implementation will look like this.

I have tried to make small meaningful commits instead of one big one, to be able to see the evolution of the code. (even knowing that in the end there will be squashed).
First time using Conventional Commit.

And last but not least, yes, the names I have chosen are probably not the best, so I am totally open to changing them if there are suggestions.

PR #639, opened 2022-11-02
feat(helm): allow users to define an external secrets for tokens

### LINKED ISSUE #638: feat(helm): allow an user defined existing secret
Hey there,

We are using Trivy for a long time in our CI, but we recently decided to give a try to the Trivy Operator.
As I noted for trivy in https://github.com/aquasecurity/trivy/issues/2586, all secrets for trivy-operator must be defined as values in order to get consumed by the secret (see https://github.com/aquasecurity/trivy-operator/blob/v0.5.0/deploy/helm/templates/config.yaml#L134). 
It would be nice to have a way to define a secret created outside of the chart (we're using external-secrets but it could be something else).

I will submit a PR on this shortly since I haven't seen any.

Let me know if it was on purpose, I'll close.

cebidhem:
## Description

With this PR, we can set `trivy.gitHubToken`, `trivy.serverToken` and `trivy.serverCustomHeaders` within a secret defined outside of the chart, by using `trivy.existingSecret` value.
This is useful to avoid having those in plain text as values, notably for those who are using solutions like external-secrets for example.

## Related issues
- Close #638

## Checklist
- [n/a] I've added tests that prove my fix is effective or that my feature works.
- [n/a] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.6.0
+version: 0.5.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -276,7 +276,7 @@ trivy:
 
   # existingSecret if an existing secret has been created outside the chart.
   # Overrides gitHubToken, serverToken, serverCustomHeaders
-  # existingSecret: ""
+  # existingSecret: "my-trivy-secret"
 
   dbRepository: "ghcr.io/aquasecurity/trivy-db"
 
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -274,9 +274,11 @@ trivy:
   #
   # serverCustomHeaders: "foo=bar"
 
-  # existingSecret if an existing secret has been created outside the chart.
+  # existingSecret if a secret containing gitHubToken, serverToken or serverCustomHeaders has been created outside the chart (e.g external-secrets, sops, etc...).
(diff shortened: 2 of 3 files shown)


deploy/helm/values.yaml
```diff
@@ -274,6 +274,10 @@ trivy:
   #
   # serverCustomHeaders: "foo=bar"
 
+  # existingSecret if an existing secret has been created outside the chart.
+  # Overrides gitHubToken, serverToken, serverCustomHeaders
+  # existingSecret: ""
```
chen-keinan, line 279: can you put an example of what are the potential values?

deploy/helm/Chart.yaml
```diff
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
```
chen-keinan, line 9: no need to change the chart version , we are automatically bumping it before we do a release

chen-keinan (changes requested):

chen-keinan:
@cebidhem thank you for the contribution 🚀 I have added couple of comments

cebidhem:
@chen-keinan I resolved the comments, thanks for the review.

However taking a deeper look into the charts, I actually have a question before pushing a new commit: how are the `Secret` and `ConfigMaps` supposed to be consumed by the `Deployment` ? I would expect to have a `envFrom:` instruction within the `deployment.yaml` file.

Anyway, let me know if I missed something, if not I'll push it.

chen-keinan:
> @chen-keinan I resolved the comments, thanks for the review.
> 
> However taking a deeper look into the charts, I actually have a question before pushing a new commit: how are the `Secret` and `ConfigMaps` supposed to be consumed by the `Deployment` ? I would expect to have either a `envFrom:` instruction within the `deployment.yaml` file for both objects, or mounted files for the properties to be used.
> 
> Anyway, let me know if I missed something, if not I'll push it.

`trivy-operator` expect to find a `Secret` called `trivy-operator-trivy-config` in `trivy-operator` namespace (by default: `trivy-system`)

the operator read the `Secret` from the cluster (`trivy-operator` namespace). not `EnvVar`

cebidhem:
Thanks for the quick reply. That means there's actually more that needs to be done in this MR. 

Knowing this, it means the user doesn't have the choice of naming its own secret: it must be called `trivy-operator-trivy-config`. Which is ok, but has to be documented in the `values.yaml`

deploy/helm/values.yaml
```diff
@@ -274,6 +274,12 @@ trivy:
   #
   # serverCustomHeaders: "foo=bar"
 
+  # existingSecret if a secret containing gitHubToken, serverToken or serverCustomHeaders has been created outside the chart (e.g external-secrets, sops, etc...).
```
chen-keinan, line 277: since it a helm setting maybe its the best to put it under [operator](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/values.yaml#L25)

chen-keinan:
@cebidhem looks OK , Added last comment

chen-keinan:
LGTM 🚀

chen-keinan (approved):

chen-keinan (approved):

PR #661, opened 2022-11-08
fix Dockerfile to use numeric user

### LINKED ISSUE #573: user should be specified numerically in Dockerfile
[currently](https://github.com/aquasecurity/trivy-operator/blob/5b9498d9d9fe186fb4861e69b161c9af87d61b38/build/trivy-operator/Dockerfile#L7) specified by username.

When i deploy in my AKS cluster, i get the following error 
```
Error: container has runAsNonRoot and image has non-numeric user (trivyoperator), cannot verify user is non-root
```

I think it can be worked around on my deploy-side by adding a RunAsUser to the securityContext, but with this simple one-line change that would not be needed.

USER trivyoperator -> USER 10000

The UID is already hardcoded to 10000 by the previous adduser statement

jrhunger:
specify user numerically in Dockerfile

## Description
Simple change to Dockerfile so that the user is specified numerically.  This allows container orchestrators to know the container will run as non-root without inspecting the passwd file in the image (which most of them can't/won't do).

## Related issues
- Close #573

## Checklist

### DIFF
--- a/build/trivy-operator/Dockerfile
+++ b/build/trivy-operator/Dockerfile
@@ -4,6 +4,6 @@ RUN adduser -u 10000 -D -g '' trivyoperator trivyoperator
 
 COPY trivy-operator /usr/local/bin/trivy-operator
 
-USER trivyoperator
+USER 10000
 
 ENTRYPOINT ["trivy-operator"]


chen-keinan:
lgtm 🚀

PR #691, opened 2022-11-15
fix(helm): fixes the install failure when using an existing trivy-operator-trivy-config secret

### LINKED ISSUE #690: helm: helm install fails when using an existingSecret

When an existing config secret exists in the namespace of trivy operator, Helm install fails with the following message:
```
Helm install failed: rendered manifests contain a resource that already exists. Unable to continue with install: Secret "trivy-operator-trivy-config" in namespace "trivy-operator" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "trivy-operator"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "trivy-operator"
```

I expect the helm release to not try to create the secret `trivy-operator-trivy-config` produced here https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/templates/config.yaml#L146

This is actually my bad, as I introduced the feature. After the review, I missed to change the line to reflect the move of the value and haven't rerun the test after the last commit...

I'll submit a PR shortly.

cebidhem:
## Description
The value that controls the creation of the trivy-operator-trivy-config secret was wrong. This fixes it.

After review, I moved the value from the trivy to the operator section, without readdressing it in the config.yaml file

This aims at fixing this.

[Gist: Helm rendered with defaults](https://gist.github.com/cebidhem/378ea3d9ee6e7e23f1d7737716ab4a90)

[Gist: Helm rendered with existing secret](https://gist.github.com/cebidhem/ecb461be3f0588e71b7fd5778e03a0a3)

## Related issues
- Close #690 

## Checklist
- [-] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [-] I've added usage information (if the PR introduces new options)
- [-] I've included a "before" and "after" example to the description (if the PR is a user interface change).

### DIFF
--- a/deploy/helm/templates/config.yaml
+++ b/deploy/helm/templates/config.yaml
@@ -140,7 +140,7 @@ data:
     {{- end }}
   {{- end }}
 ---
-{{- if not $.Values.operator.existingSecret }}
+{{- if not .existingSecret }}
 apiVersion: v1
 kind: Secret
 metadata:
--- a/deploy/helm/templates/config.yaml
+++ b/deploy/helm/templates/config.yaml
@@ -140,7 +140,7 @@ data:
     {{- end }}
   {{- end }}
 ---
-{{- if not .existingSecret }}
+{{- if not $.Values.operator.existingSecret }}
 apiVersion: v1
 kind: Secret
 metadata:
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -87,12 +87,6 @@ operator:
   # privateRegistryScanSecretsNames is map of namespace:secrets which can be used to authenticate in private registries in case if there no imagePullSecrets provided
   privateRegistryScanSecretsNames: {}
 
-  # existingSecret if a secret containing gitHubToken, serverToken or serverCustomHeaders has been created outside the chart (e.g external-secrets, sops, etc...).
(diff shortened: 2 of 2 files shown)


chen-keinan:
@cebidhem good catch , I think it was my mistake to suggest putting `existingSecret` under `.operator` , since its `trivy-config` Secret, I think the best would be to put it back as initially proposed under `.trivy`

cebidhem:
Thanks for the review @chen-keinan ! Change is done.

btw, do you think we would be able to have a fix release for this chart shortly after merge ?

chen-keinan:
> Thanks for the review @chen-keinan ! Change is done.
> 
> btw, do you think we would be able to have a fix release for this chart shortly after merge ?

I'm planning to release it officially next week as already it include some heavy changes merged to master.
I can cut `v0.7.0-rc` release tomorrow if it will help you.

cebidhem:
> I can cut v0.7.0-rc release tomorrow if it will help you.

That would be nice indeed !

Also I'll probably open a follow-up issue to discuss a different release process for the helm chart and for the application itself.

chen-keinan (approved):

PR #692, opened 2022-11-15
feat: automatic in cluster trivy-server

### LINKED ISSUE #416: Creation of automatically deploying server when using Client/Server mode

chen-keinan:

## Description
automatic in cluster trivy-server

## Related issues
- Close #416

## Checklist


bgoareguer:
You are deploying the Trivy server and its service via standalone YAML (https://github.com/chen-keinan/trivy-operator/blob/feat/automatic-deploy-trivy-server/deploy/helm/templates/trivy-server.yaml). These YAML manifests are maintained in the trivy repository, so you are duplicating code (and its maintenance).

Instead the Trivy server could be deployed via its Helm chart by including it as a dependency of the trivy-operator chart (using [conditions](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags)):
```yaml
dependencies:
  - name: trivy
    version: ~0.4.17
    repository: "https://aquasecurity.github.io/helm-charts/"
    condition: builtInTrivyServer
```

What do you think?

chen-keinan:
> You are deploying the Trivy server and its service via standalone YAML (https://github.com/chen-keinan/trivy-operator/blob/feat/automatic-deploy-trivy-server/deploy/helm/templates/trivy-server.yaml). These YAML manifests are maintained in the trivy repository, so you are duplicating code (and its maintenance).
> 
> Instead the Trivy server could be deployed via its Helm chart by including it as a dependency of the trivy-operator chart (using [conditions](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags)):
> 
> ```yaml
> dependencies:
>   - name: trivy
>     version: ~0.4.17
>     repository: "https://aquasecurity.github.io/helm-charts/"
>     condition: builtInTrivyServer
> ```
> 
> What do you think?

thanks for your feedback , I thought about it at the beginning but I preferred not to do so as:
- trivy helm chart is not maintained (not sure I want to have this coupling) as it used by community and have different needs
- it create too many resources which are not needed for operator in cluster use case
- config wise , we can reuse many relevant client/server params (which exist today) with server and vise versa (not need to have 2 diff config files its already complex today)
- The trivyServer addition (in trivy-operator) is slim and include only what we need (StatefulSet and service)

PR #771, opened 2022-12-09
feat: Optional install of CRDs

### LINKED ISSUE #723: Choose to Install CRDs
Hello, 

i would like to request the possibility to install or don't install CRDs. 
Similar to this [External Secrets Operator](https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/values.yaml#L13), it would be nice to choose if the CRD are installed or not.

alexandersperling:
* Add condition for optional install of CRDs
* Update README

## Related issues
- Close #723

## Checklist

### DIFF
--- a/deploy/crd/aquasecurity.github.io_clusterconfigauditreports.yaml
+++ b/deploy/crd/aquasecurity.github.io_clusterconfigauditreports.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.installCRDs }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -171,3 +172,4 @@ spec:
     served: true
     storage: true
     subresources: {}
+{{- end }}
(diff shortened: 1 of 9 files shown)


chen-keinan:
@alexandersperling thank you for the contribution.
I'm not sure if this change will work , there are few things need to be taken under consideration
- crd generation are controlled by [kube-builder](https://github.com/aquasecurity/trivy-operator/blob/main/Makefile#L149) , any change will be overwritten after 
- helm values has no impact crds as its out of template folder
- after changing change its best to call `make generate` making sure resources are created correctly

maybe need to think about other strategy due to the nature of crds in helm

NissesSenap:
@alexandersperling this shouldn't be needed, helm have extra logic around CRD:s built in.

When doing your helm install you can add the flag `--skip-crds if set, no CRDs will be installed. By default, CRDs are installed if not already present`

Hope that it solves any issues you might have.

alexandersperling:
True point, I think we can close this then.

PR #794, opened 2022-12-15
Support rootfs subcommand

### LINKED ISSUE #793: Support rootfs sub-command
Trivy-operator already supports the image and filesystem subcommands.

The image subcommand requires to pull the images to scan from the registries (which should not be required since the images are already in the local cache of each K8S node). Using the filesystem subcommand does not require to pull the images but it does not scan the same packages (https://aquasecurity.github.io/trivy/v0.35/docs/vulnerability/detection/language/).

I think it would make sense to add support for the rootfs subcommand so that we can have the same result as the image subcommand but without pulling the images from the registries.

bgoareguer:
## Related issues
- Close #793

## Checklist


chen-keinan:
@bgoareguer thank you for this contribution.
I wonder,  `rootfs` and `fs` works exactly the same,  setting , env variables, scan packages on fs and command syntax. 
the only different is the subcommand `rootfs` vs  `fs` would it make sense to reuse the `fs` code and just replace the sub command based on helm values, wdyt?

chen-keinan:
LGTM 🚀

chen-keinan (approved):

PR #842, opened 2023-01-11
Fixing a small bug in the grafana tutorial 

AnaisUrlichs:
The Prometheus Helm Chart was installed incorrectly (-- I must have forgotten to convert the value back after playing around with it.)


chen-keinan:
@AnaisUrlichs could you please rebase with upstream

PR #870, opened 2023-01-23
Expose trivy --ignore-policy feature

choffmeister:
## Description

This PR allows to specify a [Open Agency Policy](https://aquasecurity.github.io/trivy/v0.36/docs/vulnerability/examples/filter/#by-open-policy-agent) for running trivy. It is implemented to be as streamlined with the ignore file as possible. To specify a policy one has to extend the config like:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: trivy-operator-trivy-config
  namespace: trivy-system
data:
  trivy.ignorePolicy.application.my-app-*: |
    # applies to all workloads in namespace "application" with the name pattern "my-app-*"
  trivy.ignorePolicy.kube-system: |
    # applies to all workloads in namespace "kube-system"
  trivy.ignorePolicy: |
    # applies to all other workloads
```

## Checklist
- [X] I've added tests that prove my fix is effective or that my feature works.


chen-keinan:
@choffmeister Thanks you for raising this issue , couple of questions, if I understand you problem correctly. 
1. you wish to have some kind of policy with roles and produce vulnerability report only incase rules are matches ?
2. do you think this is relevant only for vulnerability reports ?

choffmeister:
@chen-keinan Hi. Yes, that is the key idea: Without having to list every single CVE to ignore, just describe what to mark as "mitigated" by some policies. IMHO this makes mostly sense for vulnerabilities.

I am still not really sure where the config for these mitigation policies would belong:

1. In the general trivy-operator-trivy-config conigmap, or
2. in the annotations of the workload.

The annotations of the workload have the drawback, that one might not be able to easily set rules for static pods like the ones created by kubernetes itself.

Also if the yaml structure is nice like it is using CEL or if a expression less structure is better:

1. ```yaml
    - message: Mitigation message
      matcher: |
        RegistryServer == 'registry.k8s.io' &&
        (ArtifactRepository == 'etcd' || ArtifactRepository == 'kube-apiserver') &&
        CVSSAV == 'N'
      ```
2. ```yaml
    - message: Mitigation message
      matcher:
        and:
          - registryServer: registry.k8s.io
          - or:
            - artifactRepository: etcd
            - artifactRepository: kube-apiserver
          - cvss:
              av: N
    ```

chen-keinan:
> I am still not really sure where the config for these mitigation policies would belong:
> 
> 1. In the general trivy-operator-trivy-config conigmap, or
> 2. in the annotations of the workload.
> 
what do you think about adding `policy` as a custom resource (CRD)? as a centralized place,  it will be easier to manage and more structured

choffmeister:
Oh, I did not know this feature exists. It is very close to what I want, except for the fact that this is fully ignoring the vuln instead of keeping but marking it. The rational behind the mitigation for us is, that we still want to be aware of existing but mitigated vulns to fix them eventually.

I will take a look into it and see if I can reuse the policies that already exist and extend them with "mitigate" in addition to "ignore"

chen-keinan:
> I will take a look into it and see if I can reuse the policies that already exist and extend them with "mitigate" in addition to "ignore"

Let me know if it can't be used , maybe we could think on extending it functionality

choffmeister:
@chen-keinan I have reworked the full PR to just expose the `--ignore-policy` options from trivy. The mitigation marking would be nicer, but ignoring on a policy based as offered by trivy itself is already a good start. Let me know what you think

choffmeister:
@chen-keinan One more thought on this: Most of the time one might want to specify different policies depending on what container image is scanned (or what workload/container). So picking the policy file should probably allow for this. My first naive idea would be:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: trivy-operator-trivy-config
  namespace: trivy-system
data:
  trivy.ignorePolicy: "..." # if available and no other more specific key is available, pick this
  trivy.ignorePolicy/kube-system: "..." # applies for all workloads from "kube-system" namespace
  trivy.ignorePolicy/kube-system/coredns: "..." # applies for all containers from "kube-system/coredns" workload
  trivy.ignorePolicy/kube-system/coredns/container: "..." # applies for the container "container" from "kube-system/coredns" workload
```

Good idea?

chen-keinan:
>   trivy.ignorePolicy: "..." # if available and no other more specific key is available, pick this
>   trivy.ignorePolicy/kube-system: "..." # applies for all workloads from "kube-system" namespace
>   trivy.ignorePolicy/kube-system/coredns: "..." # applies for all containers from "kube-system/coredns" workload
>   trivy.ignorePolicy/kube-system/coredns/container: "..." # applies for the container "container" from "kube-system/coredns" workload
> ```
> 
> Good idea?

I like it, as its provide the operator some flexibility, some use case thoughts:
1. assuming there are three policies define as in example above , meaning it will fallback from bottom to top incase it doesn't match.  example :
- if I have pod with `namesapce=default` and `name=nginx` then  `trivy.ignorePolicy` will be picked up.
- if I have pod with `namespace=kube-system` and `name=nginx'  then `trivy.ignorePolicy/kube-system` and etc

wdyt ?

choffmeister:
> 1. assuming there are three policies define as in example above , meaning it will fallback from bottom to top incase it doesn't match.  example :
> 
> * pod with `namesapce=default` and  `name=nginx` will pick up  `trivy.ignorePolicy`
> * pod with `namespace=kube-system` and `name=nginx`  will pick up  `trivy.ignorePolicy/kube-system`
> * pod with `namespace=kube-system` and `name=coredns`  will pick up  `trivy.ignorePolicy/kube-system/coredns`

Yes great idea. I have implemented it that way. Had to replace `/` by `.` though, as configmap keys must not container slashes. But this is fine, because workload names must not contain dots, so this naming schema is still unambiguous.

pkg/plugins/trivy/plugin_test.go
```diff
@@ -5498,6 +5798,86 @@ func TestGetContainers(t *testing.T) {
...
+			name: "fallback other-namespace other-workload",
+			configData: map[string]string{
+				"other":                              "",
+				"trivy.ignorePolicy":                 "",
+				"trivy.ignorePolicy.namespace-other": "",
+				"trivy.ignorePolicy.namespace-other.name-other-*": "",
+			},
+			expectedKey: "trivy.ignorePolicy",
+		},
+	}
```
choffmeister, line 5867: @chen-keinan These test cases show how the matching works and what as precedence over what.

pkg/plugins/trivy/plugin.go
```diff
@@ -605,6 +626,30 @@ func (p *plugin) getPodSpecForStandaloneMode(ctx trivyoperator.PluginContext, co
...
+				},
+			},
+		})
+
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      ignorePolicyVolumeName,
+			MountPath: "/etc/trivy/policy.rego",
+			SubPath:   "policy.rego",
+		})
+	}
```
chen-keinan, line 652: maybe this code be exported to function and reused in client server mode and other

docs/docs/vulnerability-scanning/trivy.md
```diff
@@ -89,6 +89,9 @@ EOF
...
+| `trivy.ignorePolicy.{ns}`             | N/A                                | It specifies a namespace specific [policy](https://aquasecurity.github.io/trivy/latest/docs/vulnerability/examples/filter/#by-open-policy-agent) file which allows to customize whicht vulnerabilities are reported by Trivy.                                                                                                                                                         |
+| `trivy.ignorePolicy.{ns}.{wl}`        | N/A                                | It specifies a namespace/workload specific [policy](https://aquasecurity.github.io/trivy/latest/docs/vulnerability/examples/filter/#by-open-policy-agent) file which allows to customize whicht vulnerabilities are reported by Trivy.                                                                                                                                                |
```
chen-keinan, line 94: typo `whicht` -- > which

also could you please update[ helm docs](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/README.md) (for artifacthub)

chen-keinan:
@choffmeister looks very good 🚀  I have added couple of comments

docs/docs/vulnerability-scanning/trivy.md
choffmeister: @chen-keinan Helm docs updated as well as the templates

pkg/plugins/trivy/plugin.go
choffmeister: @chen-keinan Done. I also did this refactor for the volumes of the .trivyignore file. In that I noticed, that there were two cases:

1. In two cases it was mounted to `/etc/trivy/.trivyignore` (I assume this is the intended location)
2. In one case it was mounted to `/tmp/trivy/.trivyignore` (I assume this was just a typo)

Now the files always go to `/etc/trivy/.trivyignore` (`/etc/trivy/policy.rego` respectively)

choffmeister:
Not sure why the test fail (https://github.com/aquasecurity/trivy-operator/actions/runs/4006579177/jobs/6891554166). Locally `make test` works fine

chen-keinan:
> Not sure why the test fail (https://github.com/aquasecurity/trivy-operator/actions/runs/4006579177/jobs/6891554166). Locally `make test` works fine

@choffmeister please rebase with upstream its a timeout issue that has been fixed.

choffmeister:
> @choffmeister please rebase with upstream its a `timeout issue` that has been fixed.

Done

choffmeister:
From my side this PR would be good to be merged.

chen-keinan:
@choffmeister could you please add example how you use it with `helm install ..`  command

choffmeister:
> helm install

Yes can do, but I have some questions on what exactly you'd like to see:

1. `deploy/helm/README.md` or `docs/gettings-started/installation/helm.md` or both?
2. And do you mean to give an example like `helm install ... --set="trivy.ignorePolicy=..."`? Because the policy is a multi line string and not really feasible for passing as CLI mark.
3. Or do you mean adding a commented out example into the helms `values.yaml`?

choffmeister:
@chen-keinan Let me know if https://github.com/aquasecurity/trivy-operator/pull/870/commits/d88ec42bca13f2c51ca3dae4640b0998d150c569 is what you expected.

chen-keinan:
> @chen-keinan Let me know if [d88ec42](https://github.com/aquasecurity/trivy-operator/commit/d88ec42bca13f2c51ca3dae4640b0998d150c569) is what you expected.

I was thinking on adding `policies` empty `{}` by default under `trivy` in values.yaml
for example :
```yaml
trivy:
  policies: {ignorePolicy.kube-system:
      "package trivy
      default ignore = false
      ignore_severities := {\"LOW\", \"MEDIUM\"}
      ignore {
        input.Severity == ignore_severities[_]
      }"}
```

and in config.yaml will change slightly like this : 
```
{{- range $k, $v := .Values.trivy.policies }}
  {{- if hasPrefix "ignorePolicy" $k }}
  trivy.{{- $k }}: |
{{- $v | trim | nindent 4 }}
  {{- end }}
{{- end }}
```
wdyt ?

choffmeister:
> wdyt ?

I was thinking the same as well first, but for me it would have looked like this:

```yaml
trivy:
  ignorePolicies:
    namespace.workload: |
      # ...
    namespace: |
      # ...
    "": |
      # ...
```

But I had no good idea on how to handle the fallback case in that. `""` looks ugly and another well-known key like `_default` or `_fallback` must always start with a `_` to differentiate it from the namespace `default` or `fallback`.

Your idea to have `trivy.policies.ignorePolicy[...]` also does not look too nice to me because of the duplication of the policy word. IMHO how I did it at the moment I like the best - also because it is completely streamlined with the non-helm case. But if you want it change, I will not heavily fight against it.

chen-keinan:
> 
> Your idea to have `trivy.policies.ignorePolicy[...]` also does not look too nice to me because of the duplication of the policy > word. IMHO how I did it at the moment I like the best - also because it is completely streamlined with the non-helm case. > > But if you want it change, I will not heavily fight against it.
>
I see your point , I'm happy to move forward with current implementation , initial testing looks good.
I need to test few more cases and then it is good for merge 🚀

chen-keinan:
@choffmeister thank you for the contribution!  LGTM 🚀

chen-keinan (approved):

PR #1000, opened 2023-02-22
add trivy_exposedsecrets_info metric with discovered secrets details

### LINKED ISSUE #998: Additional labels for trivy_image_exposedsecrets metric
Hello folks 👋🏼 

I was looking at working on a PR to add additional labels to the `trivy_image_exposedsecrets` metrics, specifically we would like to be able to identify better the secrets that were found and use it in our alerting and dashboards.

I wanted to reach to see if this is something you'd be interested in?

I'd like to add these additional labels:
- secret_target
- secret_rule_id
- secret_category
- secret_title
(perhaps also secret_match in case the above may produce non unique results?)

I'm also debating if these should be added to the existing metric, or perhaps add a separate (optional?) metric like in the case of `trivy_vulnerability_id`? Personally, I don't think this is needed but would appreciate some guidance here.

Thanks!

/cc @chen-keinan

dkulchinsky:
## Description

Adding a new metric: `trivy_exposedsecrets_info` that provides details about discovered secrets

The metric exposes the following labels (in addition to the base set):
- secret_category
- secret_rule_id
- secret_target
- secret_title

## Related issues
- Close #998

## Checklist


dkulchinsky:
note I didn't change anything in `Chart.yaml` which I assume would be taken care of when a new release is prepped, but the chart was updated to support the new configuration flag and env variable.

dkulchinsky:
Hey @chen-keinan 👋 this is ready for review 🙏🏼

chen-keinan:
@dkulchinsky thanks , will review it later on

pkg/operator/etc/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	MetricsBindAddress                           string         `env:"OPERATOR_METRICS_BIND_ADDRESS" envDefault:":8080"`
 	MetricsFindingsEnabled                       bool           `env:"OPERATOR_METRICS_FINDINGS_ENABLED" envDefault:"true"`
 	MetricsVulnerabilityId                       bool           `env:"OPERATOR_METRICS_VULN_ID_ENABLED" envDefault:"true"`
+	MetricsExposedSecretInfo                     bool           `env:"OPERATOR_METRICS_EXPOSED_SECRET_INFO_ENABLED" envDefault:"true"`
```
chen-keinan, line 28: `MetricsExposedSecretInfo` should be false (optional) by default right ?

chen-keinan:
@dkulchinsky  pr is looking good , I have added one comment

pkg/operator/etc/config.go
dkulchinsky: ahh, good point 👍🏼 

I copied over `MetricsVulnerabilityId` which looks like is set to default true as well, should I change it to false as well?

chen-keinan (approved):

chen-keinan:
lgtm 🚀

PR #1015, opened 2023-03-02
fix!: Helm chart: add missing quote filter in config template

mbreit:
## Description

The values for `trivy.resources` were not quoted.
If the Helm value is a string containing only a number, e.g. `trivy.resources.requests.cpu: "2"`, the generated YAML for the config map is `trivy.resources.requests.cpu: 2`, which is of the type number, not string, which is not a valid value for a Kubernetes config map.
Adding the `| quote` filter should fix this bug.

## Checklist

### DIFF
--- a/deploy/helm/templates/config.yaml
+++ b/deploy/helm/templates/config.yaml
@@ -145,18 +145,18 @@ data:
   {{- with .Values.trivy.resources }}
     {{- with .requests }}
       {{- if .cpu }}
-  trivy.resources.requests.cpu: {{ .cpu }}
+  trivy.resources.requests.cpu: {{ .cpu | quote }}
       {{- end }}
       {{- if hasKey . "memory" }}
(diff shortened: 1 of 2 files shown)


chen-keinan:
@mbreit Thank you for the contribution. in order to fix the verify error , you'll need to run from project folder:
```shell
make generate-all
```
then the following file will be changed :
`deploy/static/trivy-operator.yaml`

you'll need to commit and push it

mbreit:
@chen-keinan  thanks for the quick response and the hint.

I did `make generate-all` and pushed the changes in `deploy/static/trivy-operator.yaml`.

Anything else that I can/should do?

chen-keinan:
> 
> Anything else that I can/should do?

all good now. thanks.

chen-keinan:
@mbreit lgtm 🚀

PR #1066, opened 2023-03-15
fix: fixing auth issues when accessing private registries using mirroring

### LINKED ISSUE #1064: Wrong AWS region is used when defining a Mirror for an ECR registry repository

I'm trying to use the mirroring capability of the trivy operator to map an ECR registry from one region to another, the reason for that is because I'm using [ECR private image replication](https://docs.aws.amazon.com/AmazonECR/latest/userguide/replication.html), which allows us to pull the image from the same region of the Kubernetes cluster and save with outbound traffic on ECR.

I've deployed the operator with the following helm values:
```yaml
trivy:
  registry:
    mirror:
      "000000000000.dkr.ecr.us-east-1.amazonaws.com": "000000000000.dkr.ecr.eu-west-1.amazonaws.com"
```

And then I've created a deployment with a pod spec as following:
```yaml
containers:
  - name: some-container
    image: 000000000000.dkr.ecr.us-east-1.amazonaws.com/my-app:1.0.0
```

I would expect the mirror to work as intended, and the ECR image to be pulled from the `eu-west-1` region. The mapping of the mirroring is working fine and the  image URL is being changed to the desired region, however the `AWS_REGION` environment variable on the scan job spec was incorrectly defined as `eu-east-1` instead of `eu-west-1`, which breaks the AWS ECR auth performed by trivy, 

Looking into the source code, the issue is because the `CheckAwsEcrPrivateRegistry` function is being called [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/plugins/trivy/plugin.go#L797) before the mirroring logic [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/plugins/trivy/plugin.go#L847), meaning the ECR region logic is looking to the original `c.Image` instead of looking for the `optionalMirroredImage`

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): `0.12.1`
- Kubernetes version (use `kubectl version`): `EKS 1.24`
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 13.2.1

### LINKED ISSUE #1065: Using "privateRegistryScanSecretsNames" along repository mirror is not working

Let's image that I want to use the fourth option to access private registries as documented [here](https://aquasecurity.github.io/trivy-operator/v0.12.1/tutorials/private-registries/#fourth-option-define-secrets-through-trivy-operator-configuration).

So, I would deploy the operator with the following helm values:
```yaml
operator:
  privateRegistryScanSecretsNames:
    "trivy-operator": "my-registry-auth"
trivy:
  registry:
    mirror:
      "index.docker.io": "mirror.io"
```
The `my-registry-auth` secret would be of type `kubernetes.io/dockerconfigjson` and would contain the following data in `.dockerconfigjson` key:
```json
{
  "auths": {
    "mirror.io": {
      "auth": "dXNlcjE6cGFzczEyMw=="
    }
  }
}
```
And then I've created a deployment with a pod spec as following:
```yaml
containers:
  - name: nginx
    image: nginx:1.16
```

I would expect the trivy-operator to successfully pull the image from the mirrored `mirror.io/nginx:1.16` private registry using the provide credentials in the `my-registry-auth` secret, however that is not working, the operator is correctly updating the image URL to `mirror.io/nginx:1.16`, however the credentials are not being passed to the scan job spec and the job fails to pull the image from the `mirror.io` registry.

Looking into the code, currently the credentials are being mapped per container at workload level in the `submitScanJob` function [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/vulnerabilityreport/controller/workload.go#L224-L238), however the mirror mapping is only happening afterwards at plugin `GetScanJobSpec` level [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/plugins/trivy/plugin.go#L847), so the credential is being mapped for the original registry server instead of the target mirrored repository. In my humble opinion this logic should be updated, from the workload perspective the credentials should be passed to the plugin in a raw mapping, containing the registry server and the auth credentials, then from inside the public, once the mirroring mapping is done, the logic of mapping each container can be executed.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): `0.12.1`
- Kubernetes version (use `kubectl version`): `EKS 1.24`
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 13.2.1

rodrigorfk:
## Description

This change is fixing the authentication issues when accessing private registries using mirroring. The focus here is the mirror image mapping should happen before the credentials and AWS region are discovered.

Looking into the code, currently the credentials are being mapped per container at workload level in the `submitScanJob` function [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/vulnerabilityreport/controller/workload.go#L224-L238), however the mirror mapping is only happening afterwards at plugin `GetScanJobSpec` level [here](https://github.com/aquasecurity/trivy-operator/blob/v0.12.1/pkg/plugins/trivy/plugin.go#L847), so the credential is being mapped for the original registry server instead of the target mirrored repository. In my humble opinion this logic should be updated, from the workload perspective the credentials should be passed to the plugin in a raw mapping, containing the registry server and the auth credentials, then from inside the public, once the mirroring mapping is done, the logic of mapping each container can be executed.

## Related issues
- Close #1064 
- Close #1065 

## Checklist

### DIFF
--- a/pkg/kube/secrets.go
+++ b/pkg/kube/secrets.go
@@ -13,11 +13,8 @@ import (
 
 // MapContainerNamesToDockerAuths creates the mapping from a container name to the Docker authentication
 // credentials for the specified kube.ContainerImages and image pull Secrets.
-func MapContainerNamesToDockerAuths(images ContainerImages, secrets []corev1.Secret, multiSecretSupport bool) (map[string]docker.Auth, error) {
-	auths, wildcardServers, err := MapDockerRegistryServersToAuths(secrets, multiSecretSupport)
-	if err != nil {
-		return nil, err
-	}
+func MapContainerNamesToDockerAuths(images ContainerImages, auths map[string]docker.Auth) (map[string]docker.Auth, error) {
+	wildcardServers := GetWildcardServers(auths)
 
 	mapping := make(map[string]docker.Auth)
 
@@ -53,9 +50,8 @@ func matchSubDomain(wildcardServers []string, subDomain string) string {
 
 // MapDockerRegistryServersToAuths creates the mapping from a Docker registry server
 // to the Docker authentication credentials for the specified slice of image pull Secrets.
-func MapDockerRegistryServersToAuths(imagePullSecrets []corev1.Secret, multiSecretSupport bool) (map[string]docker.Auth, []string, error) {
+func MapDockerRegistryServersToAuths(imagePullSecrets []corev1.Secret, multiSecretSupport bool) (map[string]docker.Auth, error) {
 	auths := make(map[string]docker.Auth)
-	wildcardServers := make([]string, 0)
 	for _, secret := range imagePullSecrets {
 		// Skip a deprecated secret of type "kubernetes.io/dockercfg" which contains a dockercfg file
 		// that follows the same format rules as ~/.dockercfg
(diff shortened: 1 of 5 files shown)


chen-keinan:
@rodrigorfk thanks you for the contribution , I'll be able to review it tomorrow

chen-keinan:
@rodrigorfk  I have made some tests , its look like the basic follow of [imagePullSecret](https://aquasecurity.github.io/trivy-operator/v0.12.1/tutorials/private-registries/) `second option` doesn't work on your PR , could you please check it.
I'm testing now all other options as well and I'll add tests so we have private registries covered by testing

chen-keinan:
lgtm 🚀

chen-keinan (approved):

PR #1075, opened 2023-03-20
Send webhook for deleted reports

### LINKED ISSUE #1058: Send a webhook when a report is deleted
I would like to use trivy-operator webhooks to create findings in a DefectDojo instance (via a Postee exec action).

For the moment, trivy-operator does not send any webhook when a report is deleted (e.g. when a resource is deleted from the K8S cluster).

Therefore, I can't find a solution to automatically delete DefectDojo findings related to deleted resources in the K8S cluster.

Would it be possible to send a webhook when a report (vulnerabilityreport, configauditreport, etc.) is deleted?

bgoareguer:
## Description
Before this PR, trivy-operator only sent webhooks for created/updated reports.
In this PR trivy-operator also sends webhooks for deleted reports.

To make sure the receiver of the webhook can distinguish between created/updated reports and deleted reports, the webhook has been modified so that it now contains 2 fields:
- verb (can be "update" or "delete")
- report (the actual trivy-operator report)

## Related issues
- Close #1058

## Checklist


pkg/webhook/webhookreporter.go
```diff
@@ -24,6 +25,16 @@ type WebhookReconciler struct {
...
 }
 
+const (
+	Update string = "update"
+	Delete string = "delete"
+)
+
+type WebhookBody struct {
+	Verb   string        `json:"verb"`
+	Report client.Object `json:"report"`
```
chen-keinan, line 35: maybe we should change the field name to something else (`operatorObject` or else) as the `Report` already has field called  `report` and when evaluating it , it will look like `input.report.report` look strange to me , wdyt?

chen-keinan:
@bgoareguer thanks you for the PR update , I have added couple of suggestion and `nit` comment

chen-keinan:
@bgoareguer look good 🚀  thank you for the contribution,  last thing , do you mind updating the [helm readme](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/README.md) with new param

chen-keinan:
lgtm 🚀

chen-keinan (approved):

PR #1189, opened 2023-04-25
fix: Add tolerations to trivy-server template

therapy-lf:
## Description

It is sometimes essential defining tolerations. There's a `nodeSelector` field but no tolerations settings. For us it is causing pod to be in a "Pending" state.

## Related issues
Is still not opened, seems like simple fix.


chen-keinan:
@therapy-lf thank you for the contribution lgtm 🚀 , holding merge until  trivy-operator v0.13.2 officially released

PR #1205, opened 2023-05-04
fix: pod spec hash issues when using image mirroring

### LINKED ISSUE #1204: Vulnerability reports are recreated during controller restart if Image mirroring is used

Let's image that I want to use the fourth option to access private registries as documented [here](https://aquasecurity.github.io/trivy-operator/v0.13.2/tutorials/private-registries/#fourth-option-define-secrets-through-trivy-operator-configuration).

So, I would deploy the operator with the following helm values:
```yaml
operator:
  privateRegistryScanSecretsNames:
    "trivy-operator": "my-registry-auth"
trivy:
  registry:
    mirror:
      "index.docker.io": "mirror.io"
```
The `my-registry-auth` secret would be of type `kubernetes.io/dockerconfigjson` and would contain the following data in `.dockerconfigjson` key:
```json
{
  "auths": {
    "mirror.io": {
      "auth": "dXNlcjE6cGFzczEyMw=="
    }
  }
}
```
And then I've created a deployment with a pod spec as following:
```yaml
containers:
  - name: nginx
    image: nginx:1.16
```

The operator is correctly creating a job to pull the image through the mirror, this behaviour was fixed in #1066 and since then is working fine, however there is one issue with the `podSpecHash` stored in the `VulnerabilityReports` object, instead of storing the hash for the original pod spec, the one stored is for the pod spec with the modified mirroring image, therefore, if the controller is restarted, the internal cache will be empty and the operator will assume there was a change in the pod spec because the store hash does not match with the original pod spec and will schedule another scan for the image, which is unnecessary.

This issue is only reproduced in you are using image mirroring.

**Environment:**

Trivy-Operator version (use trivy-operator version): 0.13.2
Kubernetes version (use kubectl version): EKS 1.24
OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 13.2.1

rodrigorfk:
## Description
This change is fixing the mismatch between the real and stored pod spec hash when using image mirroring, this issue was introduced in #1066 when the original container image is modified and stored in the runtime cache.

## Related issues
- Close #1204

## Checklist
- [X] I've added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/pkg/kube/resources.go
+++ b/pkg/kube/resources.go
@@ -33,6 +33,18 @@ func GetContainerImagesFromPodSpec(spec corev1.PodSpec) ContainerImages {
 	return images
 }
 
+// GetContainerImagesFromContainersList returns a map of container names
+// to container images from the specified corev1.Container array.
+func GetContainerImagesFromContainersList(containers []corev1.Container) ContainerImages {
+	images := ContainerImages{}
+
+	for _, container := range containers {
+		images[container.Name] = container.Image
+	}
+
+	return images
+}
+
 // GetContainerImagesFromJob returns a map of container names
 // to container images from the specified v1.Job.
 // The mapping is encoded as JSON value of the AnnotationContainerImages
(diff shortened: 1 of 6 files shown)


chen-keinan:
lgtm 🚀

chen-keinan (approved):

PR #1212, opened 2023-05-08
feat: certificate authority scan-job support

### LINKED ISSUE #1210: Support private certificate authority
_(Follow up of discussion https://github.com/aquasecurity/trivy/discussions/4194)_

When using Trivy against a private image registry which HTTPS certificate is signed by a private authority, the error described in https://github.com/aquasecurity/trivy/issues/757 is raised:

```
Get "https://registry.private.net/v2/": x509: certificate signed by unknown authority
```

The workaround described in https://github.com/aquasecurity/trivy/issues/757 is to disable TLS verification for the registry which can be done via Trivy Operator Helm values:
```yaml
trivy:
  insecureRegistries:
    myprivateregistry: registry.private.net
```

But it would be even better if Trivy Operator supported customizing the trusted authorities by being able to provide CA certificate through a ConfigMap for instance (this is what ArgoCD does to support private CA).

It was suggested that being able to define `SSL_CERT_DIR` env variable should be enough (which Trivy Operator does not support yet) but, if it makes sense according to you, I'd go further in providing an easy way (at least in the Helm chart) for configuring private CA without having to care about this env var directly.

chen-keinan:
## Description
certificate authority scan-job support

## Related issues
- Close #1210

## Checklist


gaeljw:
Thanks a lot @chen-keinan.

Looking at the PR, it's not entirely clear to me how is the option supposed to be used. We set it to a value which would be the path to certs on the host? And it generates automatically a volume of type hostPath?

Issue #1208, opened 2023-05-05
Webhook eventually stop sending vulnerability reports


guipal:
**What steps did you take and what happened:**

Deploy operator on cluster with a scanreportTtl set to 12h and webhookurl configured

**What did you expect to happen:**

Operator to send report through the webHook either on new reports or recreated ones.

**What really happens**

Operator eventually stop sending events for any report. This also happens for new created workloads. The error disappear when restarting the pod. 

**Anything else you would like to add:**

After some debugging it appears that operator is considering all reports as historical. Somehow the value for Ttl gets to 0s while the annotation is correct

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.13.2
- Kubernetes version (use `kubectl version`): tbc
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): eks cluster

guipal:
Additionally, I increased the log level to debug an  added information about the annotation value. Here you can see an example of the problem:
![image](https://user-images.githubusercontent.com/10268389/236771280-bf217cab-94a7-4b5e-ab1e-32f6a6dfd95c.png)
As you can see the reportTtl is 0s on the webhookreporter.go
`func (r *WebhookReconciler) reconcileReport(reportType client.Object) reconcile.Func {
	return func(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
		log := r.Logger.WithValues("report", request.NamespacedName, "ttl", reportType.GetAnnotations()[v1alpha1.TTLReportAnnotation])
		log.V(1).Info("Reconciling report")
		if ignoreHistoricalReport(reportType) {
			log.V(1).Info("Ignoring historical report")
			return ctrl.Result{}, nil
		}
		verb := Update`
In the other hand, the report annotation is set to 12h1m5s as we can see in the next screenshot.
![image](https://user-images.githubusercontent.com/10268389/236771865-7419407a-d95e-4088-b0b7-b012eab1440a.png)

Hope this can give you more insights about what can be the problem.

Thank you!

chen-keinan:
> There's no specific error in the log, but as I already mentioned operator is not even sending reports created for new deployments. (This can't be historical reports).
> 
> How do you consider a report is historical? Is there any logic or is just by setting the annotation to 0s?

when an existing deployment is changed , example :
1. deploy `kubectl apply -f https://k8s.io/examples/controllers/nginx-deployment.yaml`
[...]

PR #1216, opened 2023-05-09
Read object before compare to avoid wrong cached data

guipal:
#1208 Reading the object before checking if historical to avoid reading cached annotation.

## Related issues
- Close #XXX

## Checklist

### DIFF
--- a/pkg/webhook/webhookreporter.go
+++ b/pkg/webhook/webhookreporter.go
@@ -71,14 +71,11 @@ func (r *WebhookReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *WebhookReconciler) reconcileReport(reportType client.Object) reconcile.Func {
 	return func(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+		err := r.Client.Get(ctx, request.NamespacedName, reportType)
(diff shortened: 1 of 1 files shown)


pkg/webhook/webhookreporter.go
```diff
@@ -71,14 +71,11 @@ func (r *WebhookReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *WebhookReconciler) reconcileReport(reportType client.Object) reconcile.Func {
 	return func(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+		err := r.Client.Get(ctx, request.NamespacedName, reportType)
```
chen-keinan, line 74: error check should be done right after `client get call` , I would return `L74`
`err := r.Client.Get(ctx, request.NamespacedName, reportType)` back to L81

chen-keinan:
@guipal thank you for the contribution, I have added one comment.

pkg/webhook/webhookreporter.go
guipal: Sure, moving polishing things to keep the code more readable.

chen-keinan (changes requested):

PR #1322, opened 2023-06-23
bugfix: rbac resource with capital letter name

deven0t:
When rbac resources is having capital letter in its name then trivy-operator throws error for creating its configaudit report OR rbac report. 
This happens bcoz capital letter name is not allowed in majority of the resources except rbac. 

To resolve this, we will make sure that trivy-operator doesnot try create report with capital letter, if capital letter comes then we generate report with hashvalue name
- replace IsValidLabelValue to IsDNS1123Label which allows only lower case value and other similar constraint on length

resolve: #343


chen-keinan:
@deven0t Thanks , lgtm 🚀

chen-keinan (approved):

PR #1324, opened 2023-06-26
feat: Support scan latest revision for replicationController

deven0t:
In OCP, there is a kind called DeploymentConfig, which manages the replicationController, and as we have revisionHistory for Deployments, similarly there are revision history for Deployment Config as well. Due to which many replication contollers will be present in the cluster. We dont want to scan them as they are not active resource. So as part of this change, I have added following change
- consider latest revision scan option to limit scanning of old replication controller
- ignore system pods created from deploymentConfig


chen-keinan:
@deven0t lgtm 🚀

chen-keinan (approved):

PR #1455, opened 2023-08-23
add: preserve unknown fields in vulnerabilityreport

rndmh3ro:
## Description

I built a custom operator (https://github.com/telekom-mms/trivy-dojo-report-operator) that uses the kopf-operator (github.com/nolar/kopf/) framework. This framework can act on kubernetes-resource creation. To detect which resources were already worked on, kopf adds a the last state of the resource as an annotation or status to the resource.I cannot use the annotation-feature, because of this issue: https://github.com/kubernetes-sigs/kubebuilder/issues/2556 (annotations get too long).

So my proposal is to preserve unknown fields in the vulnerabilityreport-crd. This way I can add a status-field which kopf then can use to store the state.

## Checklist

### DIFF
--- a/deploy/helm/crds/aquasecurity.github.io_vulnerabilityreports.yaml
+++ b/deploy/helm/crds/aquasecurity.github.io_vulnerabilityreports.yaml
@@ -264,6 +264,7 @@ spec:
         required:
         - report
         type: object
+        x-kubernetes-preserve-unknown-fields: true
     served: true
     storage: true
     subresources: {}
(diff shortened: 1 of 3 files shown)


chen-keinan:
@rndmh3ro thank you for the contribution , I'm happy to add such annotation however the `validation:XPreserveUnknownFields` has been deprecated in favour of `pruning:PreserveUnknownFields` which has the same functionality, adding [docs ref](https://book.kubebuilder.io/reference/markers/crd-processing) better to use the 2nd one

chen-keinan:
lgtm 🚀

PR #2080, opened 2024-05-19
feat: filter container by regex

### LINKED ISSUE #2002: Filter by image registry or regex
**Background**
We're currently evaluating `trivy-operator` to secure our clusters regarding vulnerabilities in 3rd party container images, for example `nginx`. As we already regularly scan images we build ourselves, it would be great if we could exclude our internal AWS ECR registry from image scanning. Namespace filtering doesn't help as they contain a mix of images from different sources. Resource labelling would work, but is rather tedious to set up and ensure only the correct resources are labelled.

**Feature**
Allow filtering image _before_ scanning by registry or more broadly, by regex on the image URL, so `trivy-operator` would skip (or only include) containers running such images while considering all others.

chen-keinan:
## Description
Filter container by regex

## Related issues
- Close #2002

## Checklist

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -182,7 +182,7 @@ Keeps security report resources updated
 | trivy.vulnType | string | `nil` | vulnType can be used to tell Trivy to filter vulnerabilities by a pkg-type (library, os) |
 | trivyOperator.additionalReportLabels | string | `""` | additionalReportLabels comma-separated representation of the labels which the user wants the scanner pods to be labeled with. Example: `foo=bar,env=stage` will labeled the reports with the labels `foo: bar` and `env: stage` |
 | trivyOperator.configAuditReportsPlugin | string | `"Trivy"` | configAuditReportsPlugin the name of the plugin that generates config audit reports. |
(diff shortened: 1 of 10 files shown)


deploy/helm/README.md
```diff
@@ -182,6 +182,7 @@ Keeps security report resources updated
 | trivy.vulnType | string | `nil` | vulnType can be used to tell Trivy to filter vulnerabilities by a pkg-type (library, os) |
 | trivyOperator.additionalReportLabels | string | `""` | additionalReportLabels comma-separated representation of the labels which the user wants the scanner pods to be labeled with. Example: `foo=bar,env=stage` will labeled the reports with the labels `foo: bar` and `env: stage` |
 | trivyOperator.configAuditReportsPlugin | string | `"Trivy"` | configAuditReportsPlugin the name of the plugin that generates config audit reports. |
+| trivyOperator.exludeImages | string | `""` | exludeImages is a glob patterns for excluding images from scanning. Example: `acr.azurecr.io/*/*` will exclude images from azure container registry. |
```
ybasket, line 185: Thank you tackling this issue!

There seems to be a consistent typo throughout this PR though: ˋexludeImagesˋ should probably be called ˋexcludeImagesˋ?

chen-keinan: @ybasket thanks for the catch up

PR #2111, opened 2024-05-29
fix: Add scanJobAnnotations to the Job instead of just the Pod

### LINKED ISSUE #2109: scanJobAnnotations apply to the Pod only and not the Job

Trying to configure annotations to make botkube ignore the scanning activity. The notifications get noisy for us, and I don't need to hear about Trivy starting up all the time.

When using the `trivyOperator.scanJobAnnotations` value when configuring the operator helm chart, I would expect those annotations to apply to the Job resource, given the name.

I imagine there is a case where you'd want to configure the Pod and Job annotations separately, so maybe it makes sense to have configuration for both, but applying the `scanJobAnnotations` to the actual Job resources seems like a good step.

An alternative is we configure botkube to ignore the triv-system namespace we have Trivy installed in, but the annotations should probably work here too.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.21.1
- Kubernetes version (use `kubectl version`): 1.28.9 EKS
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): N/A

darkhelmet:
## Description

The `scanJobAnnotations` were only added to the `Pod`, and not the `Job`, despite the name. Let's add the annotations to both places.

## Related issues

- Close #2109

## Checklist


chen-keinan:
@darkhelmet run the following to resolve lint issues:
```
mage generate:all
```
```
mage generate:docs
```

chen-keinan:
@darkhelmet lgtm 🚀  thank you for the contribution

chen-keinan (approved):

PR #2161, opened 2024-06-27
feat: Provide credentials in imagePullSecret without global access

### LINKED ISSUE #2158: Configure trivy-operator to use single ImagePullSecret from operator namespace without enabling global secret access
### Discussed in https://github.com/aquasecurity/trivy-operator/discussions/2053

<div type='discussions-op-text'>

<sup>Originally posted by **maltemorgenstern** May  4, 2024</sup>
Hey there, I have a question about private registries.

We are running a private registry for all of our docker images (there are gatekeeper policies in place to enforce that only internal images can be deployed into the cluster). Therefore every deployment/namespace contains an `ImagePullSecret` with credentials to our registry.

Using the default `trivy-operator` configuration this works like a charm - and reports for all resources are being created (because the operator is using the `ImagePullSecret` from each workload to pull the image, see [the docs](https://aquasecurity.github.io/trivy-operator/latest/docs/vulnerability-scanning/private-registries/)).

The problem is that the default configuration comes with highly privileged permissions. Because `operator.accessGlobalSecretsAndServiceAccount` defaults to `true`, the deployed `ClusterRole` grants the operator access to all secrets in all namespaces - which is being criticised by our security team.

```yaml
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - create
  - get
  - update
```

Based on [this discussion](https://github.com/aquasecurity/trivy-operator/discussions/1618) instead of using the `ImagePullSecret` in each namespace one can configure the operator to use a different secret from other namespaces. In our case we would just reference the secret from the operator namespace which contains credentials that are valid for all images in the cluster:

```yaml
operator:
  # Do not allow access to everything (to increase security)
  accessGlobalSecretsAndServiceAccount: false

  # Instead configure operator to read ImagePullSecret from its own namespace
  privateRegistryScanSecretsNames: {"trivy-operator":"internal-pullsecret"}
```

But this does not work - and all scan jobs fail with `unauthorized` errors.

Looking at the code the reason seems to be that `privateRegistryScanSecretsNames` requires `accessGlobalSecretsAndServiceAccount` to be enabled?
https://github.com/aquasecurity/trivy-operator/blob/4b1c6c3aed3e115ad1e4b62c51c73dcfcfa20e92/pkg/vulnerabilityreport/controller/workload.go#L255-L269

And - in general - I can understand this requirement - because if you reference a workload namespace as the secret source that cluster-wide access is required to read the data.

But if the namespace is the operator namespace this is not needed - because the default role already contains the required permissions to read kubernetes secrets:

https://github.com/aquasecurity/trivy-operator/blob/4b1c6c3aed3e115ad1e4b62c51c73dcfcfa20e92/deploy/helm/templates/rbac/role.yaml#L18-L25

### Question

- Is there currently a way to provide access to private registries without granting the operator access to all secrets in the cluster?
- If not: what do you think about the idea to add another option for providing the credentials - which would be referencing a local ImagePullSecret and therefor not needing global access?

Kind Regard Malte</div>

maltemorgenstern:
## Description
When running a cluster that contains images from a private registry one needs to configure authentication. This is done by using kubernetes ImagePullSecrets. By default the trivy-operator is able to read the secrets attached to a target workload and use them to access the container registry.

While this is necessary when working with different registries inside the cluster, this comes with one security downside: the operator needs access to all secrets inside the cluster.

If all images are being pulled from a single private registry then one ImagePullSecret can be used for all of them. The easiest one to use is inside the operator namespace (because the operator has access to secrets in its own namespace).

This change does not impact any deployments running with default settings (global access enabled). But in case one disables that access this allows to instead read a single ImagePullSecret and use it for all images.

## Related issues
- Close #2158 

## Checklist

### DIFF
--- a/pkg/vulnerabilityreport/controller/workload.go
+++ b/pkg/vulnerabilityreport/controller/workload.go
@@ -252,19 +252,34 @@ func (r *WorkloadController) SubmitScanJob(ctx context.Context, owner client.Obj
 		"name", owner.GetName(), "namespace", owner.GetNamespace())
 	var err error
 	credentials := make(map[string]docker.Auth, 0)
+
+	privateRegistrySecrets, err := r.Config.GetPrivateRegistryScanSecretsNames()
+	if err != nil {
+		return err
+	}
+
+	pConfig, err := r.PluginContext.GetConfig()
+	if err != nil {
+		return err
+	}
+
+	multiSecretSupport := trivy.MultiSecretSupport(trivy.Config{PluginConfig: pConfig})
+
 	if r.AccessGlobalSecretsAndServiceAccount && len(reusedReports) == 0 {
-		privateRegistrySecrets, err := r.Config.GetPrivateRegistryScanSecretsNames()
+		// Global access is enabled - therefore imagePullSecrets from the podSpec can be used
+		credentials, err = r.CredentialsByServer(ctx, owner, privateRegistrySecrets, multiSecretSupport, true)
 		if err != nil {
 			return err
 		}
-		pConfig, err := r.PluginContext.GetConfig()
-		if err != nil {
-			return err
-		}
-		multiSecretSupport := trivy.MultiSecretSupport(trivy.Config{PluginConfig: pConfig})
-		credentials, err = r.CredentialsByServer(ctx, owner, privateRegistrySecrets, multiSecretSupport)
-		if err != nil {
-			return err
+	} else {
+		// Global access is disabled - check if privateRegistrySecrets references an imagePullSecret in the operator namespace
+		imagePullSecretName, ok := privateRegistrySecrets[r.Config.Namespace]
+		if ok {
+			// privateRegistrySecrets references an imagePullSecret in the operator namespace - use this for pulling private images
+			credentials, err = r.CredentialsByServer(ctx, owner, map[string]string{r.Config.Namespace: imagePullSecretName}, multiSecretSupport, false)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
--- a/pkg/kube/secrets.go
+++ b/pkg/kube/secrets.go
@@ -124,7 +124,7 @@ const (
 type SecretsReader interface {
 	ListByLocalObjectReferences(ctx context.Context, refs []corev1.LocalObjectReference, ns string) ([]corev1.Secret, error)
 	ListImagePullSecretsByPodSpec(ctx context.Context, spec corev1.PodSpec, ns string) ([]corev1.Secret, error)
-	CredentialsByServer(ctx context.Context, workload client.Object, secretsInfo map[string]string, multiSecretSupport bool) (map[string]docker.Auth, error)
+	CredentialsByServer(ctx context.Context, workload client.Object, secretsInfo map[string]string, multiSecretSupport bool, globalAccessEnabled bool) (map[string]docker.Auth, error)
 }
 
 // NewSecretsReader constructs a new SecretsReader which is using the client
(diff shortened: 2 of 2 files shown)


maltemorgenstern:
Hey, this MR is not finished (and tests will fail for sure), but I wanted to get your opinion on my approach first - afterwards, I will try to add and fix the necessary tests.

```yaml
operator:
  accessGlobalSecretsAndServiceAccount: false
  privateRegistryScanSecretsNames: {"trivy-operator":"internal-pullsecret"}
```

With this change the above configuration would allow the trivy-operator to pull all images inside the cluster that use the same private registry - while not having to grant it access to every secret in the entire cluster.

WDYT?

chen-keinan:
@maltemorgenstern looks ok in general

maltemorgenstern:
Hey @chen-keinan, I finally found the time to finalize this PR - it's ready for review now.

FYI: I don't have much experience in Go - so any hints about styling or optimizations are welcome!

maltemorgenstern:
Hey @chen-keinan, just wanted to check in - what to you think about these changes?
Thanks!

maltemorgenstern:
Hey, we would still like this to be merged. @chen-keinan do you mind giving it a review?

simar7:
@afdesk could you take a look when you have a chance?

deploy/static/trivy-operator.yaml
```diff
@@ -3169,6 +3169,25 @@ spec:
...
 ---
+# Source: trivy-operator/templates/rbac/clusterrolebinding.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: trivy-operator
+  labels:
+    app.kubernetes.io/name: trivy-operator
+    app.kubernetes.io/instance: trivy-operator
+    app.kubernetes.io/version: "0.21.3"
```
simar7, line 3180:
```suggestion
    app.kubernetes.io/version: "0.23.0"
```

deploy/static/trivy-operator.yaml
```diff
@@ -3537,20 +3556,21 @@ rules:
...
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
-  name: trivy-operator
+  name: trivy-operator-leader-election
+  namespace: trivy-system
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
     app.kubernetes.io/version: "0.21.3"
```
simar7, line 3568:
```suggestion
    app.kubernetes.io/version: "0.23.0"
```

afdesk:
@maltemorgenstern 
thanks a lot for your contribution and sorry for really long delay to your PR.

can we follow up this job? thanks

maltemorgenstern:
Hey @afdesk and @simar7,
thanks for looking into this MR. I will try to rebase/update the branch in the next few days (just need to find some time) 👍

afdesk:
@maltemorgenstern great to hear it! thanks a lot for your contribution! 

we (@simar7 ) have a concern about backward compatibilities, please, feel free to correct me.

Now `globalAccessEnabled` is disabled by default, right? 
maybe should we keep it `true` by default?

afdesk:
@maltemorgenstern could you also rebase the PR with the main branch? thanks

pkg/vulnerabilityreport/controller/workload.go
```diff
@@ -252,19 +252,34 @@ func (r *WorkloadController) SubmitScanJob(ctx context.Context, owner client.Obj
...
-			return err
+	} else {
+		// Global access is disabled - check if privateRegistrySecrets references an imagePullSecret in the operator namespace
+		imagePullSecretName, ok := privateRegistrySecrets[r.Config.Namespace]
+		if ok {
+			// privateRegistrySecrets references an imagePullSecret in the operator namespace - use this for pulling private images
+			credentials, err = r.CredentialsByServer(ctx, owner, map[string]string{r.Config.Namespace: imagePullSecretName}, multiSecretSupport, false)
+			if err != nil {
+				return err
+			}
```
afdesk, line 282: also this `else` statement will run If `resusedReports` isn't empty

pkg/vulnerabilityreport/controller/workload.go
```diff
@@ -252,19 +252,34 @@ func (r *WorkloadController) SubmitScanJob(ctx context.Context, owner client.Obj
...
+	if err != nil {
+		return err
+	}
+
+	pConfig, err := r.PluginContext.GetConfig()
+	if err != nil {
+		return err
+	}
+
+	multiSecretSupport := trivy.MultiSecretSupport(trivy.Config{PluginConfig: pConfig})
```
afdesk: also I have a concern about this block.
should we run it for existed KBOM reports?

afdesk (changes requested):
I left some marks

pkg/vulnerabilityreport/controller/workload.go
afdesk: changed https://github.com/aquasecurity/trivy-operator/pull/2161/commits/9d514f8f5d0f922edb42f58ba9b39a6bb4ea9876

pkg/vulnerabilityreport/controller/workload.go
afdesk: refactored https://github.com/aquasecurity/trivy-operator/pull/2161/commits/9e9821884d138be8bfe013a4e5bbd1ed932218ed

afdesk:
@simar7 @maltemorgenstern 
I've updated a bit this useful PR.
could you take another look again?
thanks

PR #2252, opened 2024-09-05
fix missing server affinity

jfcoz:
## Description

Also use affinity in trivy-server statefulset


afdesk:
@jfcoz thanks for your contribution

there is an issue with `license/cla` step here.

kladiv:
Hello,
i guess _affinity_, _tolerations_, _nodeSelector_ in the statefulset manifest of Trivy Server should use `.Values.trivy.server.<spec>` instead of top level values.

PR #2286, opened 2024-10-15
fix: set `follow` to false when retrieving logs

### LINKED ISSUE #2264: needlessly using follow option when retrieving logs of completed job
Seeing error "illegal base64 data at input byte 5761" in trivy-operator logs.  
Looking at the completed job in question, we are able to successfully query and decode the logs:
```
$ kubectl -n trivy-system logs scan-vulnerabilityreport-d4856fb89-jbzzz  > x
$ wc -c x
5755 x
## note that the illegal input byte 5761 is > the size of this file 5755
$ base64 --decode < x > y
$ head -3 y
{
  "SchemaVersion": 2,
  "CreatedAt": "2024-09-17T09:43:51.613221809Z",
```
However if we add the -f option to kubectl logs, we get an extra line at the end:
```
MOo50krocvh7q08U2UkR+LzJgnsPQtfXuwggOKzv/V0EkLKRJH4Peyf9VD/NFLEkT/4x/pWf23/o
82xjGG/1/P9/+L7Ob/x3H/r/8VVPF4/7f7OLpw6dBS/1pUrrVJx/T/zuahnSGIzw5Q/wcrT7KmMw
/4u5IpwoSFKbvOGA
failed to create fsnotify watcher: too many open files
```
OK, granted this is an issue with our system, however, it also implies that trivy-operator is using the "follow" option when pulling the logs from a completed job, which shouldn't be necessary (no new log entries will be created for a completed job).

I expect that trivy-operator can retrieve the logs from a completed job even if there is an issue on the system that causes an extra non-base64 line to be added when retrieving logs in Follow mode.

Looking in the logs, i see Follow: true [in GetLogsByJobAndContainerName](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/kube/logs.go#L45)
That is only called in [pkg/configauditreport/controller/nodecollector.go](https://github.com/aquasecurity/trivy-operator/blob/18e40db60a957cf233afeb442bddca62a03e45e9/pkg/configauditreport/controller/nodecollector.go#L118) and [pkg/vulnerabilityreport/controller/scanjob.go](https://github.com/aquasecurity/trivy-operator/blob/18e40db60a957cf233afeb442bddca62a03e45e9/pkg/vulnerabilityreport/controller/scanjob.go#L239), in each case inside a function called "processCompleteScanJob" (although there is an intermediate function on the 2nd one)

So if this is always called on Completed jobs, it seems the Follow: true is not necessary.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.22.0 (but issue is there in current master)
- Kubernetes version (use `kubectl version`): 1.27.11 (but not version dependent)
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):  Oracle Linux 9.4

jrhunger:
## Description
Set Follow=false when retrieving logs from completed job.
Using Follow mode in this case is unnecessary because:
* Follow is used to wait for additional logs that may come from the container
* By the time the operator is collecting the container logs, the container has already completed and will not generate more logs

In cases where the underlying node has resources (fsnotify handle) exhaustion this causes the logs to return an error after the base64 data "failed to create fsnotify watcher: too many open files", and trivy fails to process that scan job due to "illegal base64 data".

I tested this simple change in a cluster that was having the issue, and the operator processed and deleted all the completed scan jobs that were previously being continuously re-processed with base64 failures.

## Related issues
- Close #2264 

## Checklist

### DIFF
--- a/pkg/kube/logs.go
+++ b/pkg/kube/logs.go
@@ -42,7 +42,7 @@ func (r *logsReader) GetLogsByJobAndContainerName(ctx context.Context, job *batc
 
 	return r.clientset.CoreV1().Pods(pod.Namespace).
 		GetLogs(pod.Name, &corev1.PodLogOptions{
-			Follow:    true,
+			Follow:    false,
 			Container: containerName,
 		}).Stream(ctx)
 }


jrhunger:
I have signed the CLA twice, once on mobile and once on my phone. Also clicked the recheck button multiple times.  Maybe that integration is having issues?

The integration test is timing out at 300s on "When unmanaged Pod is created [It] Should create VulnerabilityReport".  In my cluster, the modified trivy-operator creates a VulnerabilityReport for an unmanaged pod.  On the line above that ([460](https://github.com/aquasecurity/trivy-operator/actions/runs/11351581678/job/31572816163?pr=2286#step:7:461) in the test output, there is an error (below) which i think is the real cause of the failure:

failed to download policies: failed to download built-in policies: download error: oci download error: failed to fetch the layer: GET https://ghcr.io/v2/aquasecurity/trivy-checks/blobs/sha256:cba49b6781cfcdeb6b063283a711ce0ddb1f36d6e2a5db69ef7d2e3f13998149: TOOMANYREQUESTS: retry-after: 885.605µs, allowed: 44000/minute

simar7:
The PR seems fine to me but can we confirm that we only use this function to get logs for completed jobs? Otherwise we would need the ` follow` option to be set.

afdesk:
> The PR seems fine to me but can we confirm that we only use this function to get logs for completed jobs? Otherwise we would need the ` follow` option to be set.

to fix #2264 I took a look at this PR too
It seems @jrhunger is right: `GetLogsByJobAndContainerName` calls only for completed jobs

afdesk (approved):

simar7 (approved):

PR #2300, opened 2024-10-29
perf: Optimize cache by minimizing unnecessary data storage.

mjshastha:
- Updated `cache.Options` with a `DefaultTransform` function that removes managed fields and the `"kubectl.kubernetes.io/last-applied-configuration"` annotation from objects before storing them in the cache.
- Stripping managed fields reduces cache memory overhead, which improves performance in large clusters. ref: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#TransformStripManagedFields
- Removing the `"last-applied-configuration"` annotation prevents unnecessary cache bloat.

This enhancement minimizes cache size and resource usage, improving controller runtime performance and memory management.


afdesk (approved):
LGTM

PR #2353, opened 2024-12-09
fix: Missing credentials for downloading the java-db from private registry

tom1299:
## Description
The main container with the image command of a scan job fails when pulling the `java-db` from a private registry because of missing credentials. This PR introduces a second init container analogous to the one that downloads the `trivy-db` but downloads the `java-db` if it is to be pulled from a private registry.
This results in the main container not having to download the `java-db` since it is up to date.

## Checklist

### DIFF
--- a/pkg/plugins/trivy/config.go
+++ b/pkg/plugins/trivy/config.go
@@ -233,6 +233,14 @@ func (c Config) GetSkipJavaDBUpdate() bool {
 	return boolVal
 }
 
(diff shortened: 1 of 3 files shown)


tom1299:
Please note that I added an additional test case since I did not want to modify the existing ones.

afdesk (approved):

simar7 (approved):

simar7:
nice, thanks @tom1299 for your contribution!

PR #2375, opened 2025-01-14
fix: Remove job deletion for jobs with TTLSecondsAfterFinished set

### LINKED ISSUE #2362: Vulnerability ScanJob is immediately deleted
Set `operator.scanJobTTL: "5m"` in the helm chart values in order to retain the Job and Pod for 5 minutes after completion.
Instead of retaining the Job and Pod, the scan job was immediately deleted.

The Scanjob and corresponding Pod to remain 5 minutes after reaching the status `Completed`

The vulnerability controller does not honour `job.Spec.TTLSecondsAfterFinished`. It directly deletes completed jobs. See [code here](https://github.com/aquasecurity/trivy-operator/blob/a1378257d977af1d3a3f44a519150acfc94e0476/pkg/vulnerabilityreport/controller/scanjob.go#L210)
I alter the code with the following lines before the deletion which solved the problem:
```golang
if job.Spec.TTLSecondsAfterFinished != nil {
    log.V(1).Info("Job has TTL. Omitting deletion", "owner", owner)
    return nil
}
```
I can open a PR to fix this issue.

**Environment:**

- Trivy-Operator version: Custom build based on 7af084e861d2bcda689dc7dbfcbf1338efe3ddaf
- Kubernetes version: `Client Version: v1.31.0`, `Server Version: v1.30.0`
- OS: Fedora 39

tom1299:
## Description
Currently jobs are directly deleted after they are completed by the trivy-operator without honouring the jobs `ttlSecondsAfterFinished` and thus also the helm configuration parameter `operator.scanJobTTL`.
The change introduced in this PR will only delete the job if it has no `ttlSecondsAfterFinished`.

## Related issues
- Close #2362 

## Checklist

### DIFF
--- a/pkg/vulnerabilityreport/controller/scanjob_test.go
+++ b/pkg/vulnerabilityreport/controller/scanjob_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"testing"
 
-	alpha1 "github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -15,7 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
-	"github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
+	v1alpha1 "github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1"
 	"github.com/aquasecurity/trivy-operator/pkg/kube"
 	"github.com/aquasecurity/trivy-operator/pkg/operator/etc"
 	"github.com/aquasecurity/trivy-operator/pkg/trivyoperator"
@@ -43,7 +42,7 @@ func setupTestEnvironment(testid string) (*ScanJobController, *batchv1.Job, erro
 	// Create the vulnerability report for the scanned replica set with the appropriate labels
 	vulnReport := v1alpha1.VulnerabilityReport{
 		TypeMeta: metav1.TypeMeta{
-			APIVersion: alpha1.SchemeGroupVersion.String(),
+			APIVersion: v1alpha1.SchemeGroupVersion.String(),
 			Kind:       "Job",
 		},
 		ObjectMeta: metav1.ObjectMeta{


pkg/vulnerabilityreport/controller/scanjob.go
```diff
@@ -357,6 +357,10 @@ func (r *ScanJobController) completedContainers(ctx context.Context, scanJob *ba
 }
 
 func (r *ScanJobController) deleteJob(ctx context.Context, job *batchv1.Job) error {
+	if job.Spec.TTLSecondsAfterFinished != nil && *job.Spec.TTLSecondsAfterFinished != 0 {
```
simar7, line 360: I read through the docs but it wasn't clear to me why we should check for both `!=nil` and `!=0`. Could you explain that?

tom1299: Actually I just realiezed that `!=0` is incorrect. If `ttl` is 0, the job could obviously be deleted. I would remove that from the if statement.

tom1299: Actually now that you mentioned it, it seems to me that the `!=0` is superfluous / not really necessary. I would remove this from the if statement.

simar7: Could you update your PR?

tom1299: @simar7 I'm sorry but somehow I totally forgot to do the update. I made the change and squashed the commits so that the commit history looks better.

simar7:
@afdesk can you take another look?

afdesk:
@simar7 will merge it, right?

simar7:
@tom1299 is it possible to write a test for this case? While it is a trivial change, it is an important one. Maybe we can use this test setup? https://github.com/aquasecurity/trivy-operator/blob/2ed26a2ff2756458c699e2cfaee3b7130e4946b1/tests/envtest/controller_test.go#L200-L226 WDYT?

tom1299:
@simar7 Adding a test case like above is a very good idea. I will have a deeper look into the env tests and try to add a test case for the ttl jobs.

tom1299:
@simar7 So I had a look at the env tests. And the main issue for adding a test case to the env tests is that the [ScanJobController](https://github.com/aquasecurity/trivy-operator/blob/85bab6ec6b2dba4908401dd7995ef0fe73e77445/pkg/vulnerabilityreport/controller/scanjob.go#L31) can not be easily added to the env test setup [here](https://github.com/aquasecurity/trivy-operator/blob/85bab6ec6b2dba4908401dd7995ef0fe73e77445/tests/envtest/suite_test.go#L60). Since the `ScanJobController` contains the method to test, it will never be executed.
I also think that even if it would be possible to add the controller, the `JobHasAnyCondition` Predicate would prevent the Controller from processing the job since the Jobs created in the env tests do not seem to have a status.
Here is the predicate code:
```golang
var JobHasAnyCondition = predicate.NewPredicateFuncs(func(obj client.Object) bool {
	if job, ok := obj.(*batchv1.Job); ok {
		return len(job.Status.Conditions) > 0
	}
	return false
})
```
And it is registered with `ScanJobController` [here](https://github.com/aquasecurity/trivy-operator/blob/85bab6ec6b2dba4908401dd7995ef0fe73e77445/pkg/vulnerabilityreport/controller/scanjob.go#L53C85-L53C103):
```golang
func (r *ScanJobController) SetupWithManager(mgr ctrl.Manager) error {
	var predicates []predicate.Predicate
	if !r.ConfigData.VulnerabilityScanJobsInSameNamespace() {
		predicates = append(predicates, InNamespace(r.Config.Namespace))
	}
	predicates = append(predicates, ManagedByTrivyOperator, IsVulnerabilityReportScan, JobHasAnyCondition)
	return ctrl.NewControllerManagedBy(mgr).
		For(&batchv1.Job{}, builder.WithPredicates(predicates...)).
		Complete(r.reconcileJobs())
}
``` 
By the way, there seems to be a redundant implementation in the `ScanJobController` itself [here](https://github.com/aquasecurity/trivy-operator/blob/85bab6ec6b2dba4908401dd7995ef0fe73e77445/pkg/vulnerabilityreport/controller/scanjob.go#L73):
```golang
if len(job.Status.Conditions) == 0 {
	log.V(1).Info("Ignoring Job without conditions")
	return ctrl.Result{}, nil
}
```
Which means that the when the Predicate would be removed, the ScanJobController would still not process the job. For me this might also be an issue that needs to be fixed.

So I think it would be rather hard to test the changes in the env tests. Or maybe you could give me another hint how to add a test ? Otherwise I can suggest looking into the integration tests. Maybe we could add a test case here ?

tom1299:
As an alternative I added a integration test here https://github.com/aquasecurity/trivy-operator/pull/2375/commits/932e1a85309292236c0050cca0518f8c63156660. the tests validates that if ttl is set for scan jobs, they are not deleted immediately. What do you think about that ? Would that be a way to to the testing ?

afdesk:
@tom1299 could you merge the main branch here?
it should fix some failed tests
thanks

tom1299:
@afdesk The reason why `generate:verifydocs` fails is because I had to change the `OPERATOR_SCAN_JOB_TTL` and `OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT` in order to test the change. Since those are non default values, the verification fails. Is there a better way to alter the setup maybe ?

afdesk:
> Is there a better way to alter the setup maybe ?

maybe I miss something but `mage generate:docs` and `mage generate:all` should help

tom1299:
In order to test whether a scan job is not deleted if ttl is set I directly changed the configuration parameters [`OPERATOR_SCAN_JOB_TTL` and `OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT`](https://github.com/aquasecurity/trivy-operator/pull/2375/files#diff-6a6c7dca99aa76589f23c77ed3a9f61c830cd67d933e9887c321a015d373d174L2995) since the integration tests use the static deployment file for the trivy-operator.
But those are non default values as in [values.yaml](https://github.com/aquasecurity/trivy-operator/blob/7961baa247ce94d1d06fce4e5dd02f62b85829ba/deploy/helm/values.yaml#L63). I was reluctant to change those default values since that would impact all users of the trivy-operator.
`mage generate:all` will use the [values.yaml](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/values.yaml) with the script [update-static.yaml.sh](https://github.com/aquasecurity/trivy-operator/blob/main/hack/update-static.yaml.sh) to generate the static deployment file thus overriding the changes I made in order to run the new integration test successfully.
During the GitHub workflow `generate-all` is called and the `verify-generated.sh` detects the change in the static deployment yaml and fails.
I am not sure how to solve that...

deploy/static/trivy-operator.yaml
```diff
@@ -2992,9 +2992,9 @@ metadata:
     app.kubernetes.io/managed-by: kubectl
 data:
   OPERATOR_LOG_DEV_MODE: "false"
-  OPERATOR_SCAN_JOB_TTL: ""
+  OPERATOR_SCAN_JOB_TTL: "30s"
   OPERATOR_SCAN_JOB_TIMEOUT: "5m"
-  OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT: "10"
+  OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT: "20"
```
simar7, line 2997: I understand you want to use this file for tests but it is auto generated so any changes here will not persist when generation takes place.

Instead of change the defaults as that is not the point of this PR, how about we test it differently? I found this unit test that can allow the modification for TTL values and be used to test this change https://github.com/aquasecurity/trivy-operator/blob/29d14bf6c8b5473d1a8c026c574cff2fdfcf22a2/pkg/operator/ttl_report_test.go#L28

simar7:
> In order to test whether a scan job is not deleted if ttl is set I directly changed the configuration parameters [`OPERATOR_SCAN_JOB_TTL` and `OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT`](https://github.com/aquasecurity/trivy-operator/pull/2375/files#diff-6a6c7dca99aa76589f23c77ed3a9f61c830cd67d933e9887c321a015d373d174L2995) since the integration tests use the static deployment file for the trivy-operator.
> But those are non default values as in [values.yaml](https://github.com/aquasecurity/trivy-operator/blob/7961baa247ce94d1d06fce4e5dd02f62b85829ba/deploy/helm/values.yaml#L63). I was reluctant to change those default values since that would impact all users of the trivy-operator.
> `mage generate:all` will use the [values.yaml](https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/values.yaml) with the script [update-static.yaml.sh](https://github.com/aquasecurity/trivy-operator/blob/main/hack/update-static.yaml.sh) to generate the static deployment file thus overriding the changes I made in order to run the new integration test successfully.
> During the GitHub workflow `generate-all` is called and the `verify-generated.sh` detects the change in the static deployment yaml and fails.
> I am not sure how to solve that...

hi @tom1299 - I was able to get back to review this PR again see my comment regarding testing here https://github.com/aquasecurity/trivy-operator/pull/2375/files#r1966339703

deploy/static/trivy-operator.yaml
tom1299: Thanks for the review and for the hint to the ttl_report_test. I added a test similar to that one in a new file [scanjob_test.go](https://github.com/aquasecurity/trivy-operator/pull/2375/files#diff-a4718cd1e513c622d28b559ef021936c1b851483ec55919ee0786f712cb3070d). It contains a test for both scenarios; When the job should be deleted immediately and when not. So even better than the env tests.
What do you think about that ? I could delete the env test in facvor of this test then.

Rotzke:
Hi @simar7! Any updates on the PR? There is a reply from @tom1299.

The issue is also affecting my client, may I maybe help somehow to move things forward? FluentBit can't gather Kubernetes metadata from the scan job pods because they are deleted so fast.

simar7 (approved):

tom1299:
@simar7 Linting and testing actions fail for some reason which I can not really figure out. Both mage tasks work fine when I run them locally. Maybe you could help me figuring out why these tests fail on Github ?

afdesk:
@tom1299 thanks for your errofts!
 could you rebase your code with the main branch and update tests?

afdesk:
@tom1299 so sorry for long response.
I'll try to rebase your great PR with the main branch, it seems it'll be easy to re-create a new PR with the same final files. (
could you arise a new PR for it because it's a really useful fix? 
thanks a lot!

tom1299:
> @tom1299 so sorry for long response. I'll try to rebase your great PR with the main branch, it seems it'll be easy to re-create a new PR with the same final files. ( could you arise a new PR for it because it's a really useful fix? thanks a lot!

@afdesk Just so that I understand it correctly: I should create a new PR from main with the changes done in this PR ? That should not be a problem and I would be happy to do it.

afdesk:
> I should create a new PR from main with the changes done in this PR ?

yes, please!
I couldn't rebase it correctly
thanks a lot

tom1299:
Closed in favour of #2632

PR #2380, opened 2025-01-17
feat: apply gofumpt formatting

jokestax:
## Description
Formatted codebase using gofumpt -w .. No functional changes made. :)


simar7:
Thanks for the PR but we already have another lint PR in progress which should take care of things. https://github.com/aquasecurity/trivy-operator/pull/2392

PR #2392, opened 2025-01-23
chore(lint): Add support for linter in mage

simar7:
## Description

- Add support for linter in mage
- Address linting issues

`.golanci.yaml` is taken from Trivy.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,14 +1,159 @@
-linters:
-  disable:
-    - structcheck # FIXME: Currently doesn't work on Go 1.18: https://github.com/golangci/golangci-lint/issues/2649
-  enable:
-    - errorlint
-    - gocritic
-    - gosec
-    # - unparam FIXME: Currently doesn't work on Go 1.18: https://github.com/golangci/golangci-lint/issues/2649
 linters-settings:
+  depguard:
+    rules:
+      main:
+        list-mode: lax
+        deny:
+          # Cannot use gomodguard, which examines go.mod, as "golang.org/x/exp/slices" is not a module and doesn't appear in go.mod.
+          - pkg: "golang.org/x/exp/slices"
+            desc: "Use 'slices' instead"
+          - pkg: "golang.org/x/exp/maps"
+            desc: "Use 'maps' or 'github.com/samber/lo' instead"
+  dupl:
+    threshold: 100
+  errcheck:
+    check-type-assertions: true
+    check-blank: true
+  gci:
+    sections:
+      - standard
+      - default
+      - prefix(github.com/aquasecurity/)
+      - blank
+      - dot
+  goconst:
+    min-len: 3
+    min-occurrences: 3
   gocritic:
-    enabled-checks:
-      - importShadow
+    disabled-checks:
+      - appendAssign
+      - unnamedResult
+      - whyNoLint
+      - indexAlloc
+      - octalLiteral
+      - hugeParam
+      - rangeValCopy
+      - regexpSimplify
+      - sloppyReassign
+      - commentedOutCode
+    enabled-tags:
+      - diagnostic
+      - style
+      - performance
+      - experimental
+      - opinionated
+    settings:
+      ruleguard:
+        failOn: all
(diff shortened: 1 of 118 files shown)


pkg/configauditreport/controller/helper.go
```diff
@@ -46,8 +47,8 @@ func evaluate(ctx context.Context, policies *policy.Policies, resource client.Ob
 	if err != nil {
 		return Misconfiguration{}, err
 	}
-	infraChecks := make([]v1alpha1.Check, 0)
-	checks := make([]v1alpha1.Check, 0)
+	var infraChecks []v1alpha1.Check
```
afdesk, line 50: it looks like it's not equal update.

misc/lint/rules.go
```diff
@@ -0,0 +1,37 @@
...
+
+// cf. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices
+func declareEmptySlices(m dsl.Matcher) {
+	m.Match(
+		`$name := []$t{}`,
+		`$name := make([]$t, 0)`,
+	).
+		Suggest(`var $name []$t`).
+		Report(`replace '$$' with 'var $name []$t'`)
+}
```
afdesk, line 15: as for me, `var $name []$t` isn't equal `$name := make([]$t, 0)`,
because `$name := make([]$t, 0)` contains two operations:
- create a new variable 
- init this one with an empty array,

but `var $name []$t` does only creating

pkg/kube/secrets.go
```diff
@@ -142,7 +143,7 @@ type secretsReader struct {
 }
 
 func (r *secretsReader) ListByLocalObjectReferences(ctx context.Context, refs []corev1.LocalObjectReference, ns string) ([]corev1.Secret, error) {
-	secrets := make([]corev1.Secret, 0)
+	var secrets []corev1.Secret
```
afdesk, line 146: imho it's not equal too, and it's important for tests

afdesk (changes requested):

misc/lint/rules.go
afdesk: so after `var $name []$t`, `$name` is nil, but after `$name := make([]$t, 0)` '$name` is `[]$t`

.golangci.yml
```diff
@@ -1,14 +1,159 @@
+linters-settings:
+  depguard:
+    rules:
+      main:
+        list-mode: lax
+        deny:
+          # Cannot use gomodguard, which examines go.mod, as "golang.org/x/exp/slices" is not a module and doesn't appear in go.mod.
+          - pkg: "golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
simar7: thanks, removed!

misc/lint/rules.go
simar7: You're right, in this case as per the golang spec it makes sense to use `make` due to JSON being involved. I've updated it.

pkg/kube/secrets.go
simar7: updated.

simar7:
@afdesk thanks for the review, could you take another look? The tests are green now.

afdesk (approved):

afdesk:
@simar7 LGTM

I left two comments just to make sure you saw and approved these changes.

pkg/plugins/trivy/flags.go
```diff
@@ -1,20 +1,20 @@
...
 
 import (
-	"github.com/Masterminds/semver"
+	"github.com/aquasecurity/go-version/pkg/semver"
 )
 
-func compareTagVersion(currentTag string, contraint string) bool {
-	c, err := semver.NewConstraint(contraint)
+func compareTagVersion(currentTag, constraint string) bool {
+	c, err := semver.NewConstraints(constraint)
```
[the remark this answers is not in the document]
simar7: partially yes. The change to eliminate the redundant type was suggested by the linter and the typo was fixed by me.

As for the semver package, We have had issues in the past with upstream semver, for aqua projects we use the aqua go-version pkg.

PR #2398, opened 2025-01-27
perf: Exclude kube-bench pods from being evaluated in the Trivy operator

mjshastha:
In a scaled 800-node cluster, running the Trivy operator with KE results in 1600 pods being created for kube-bench. Since kube-bench is an internal Aqua component, its evaluation can be safely skipped. By excluding kube-bench from the Trivy operator's scope, we can reduce the overall memory consumption of the Trivy operator, optimizing its performance in large-scale environments.

### DIFF
--- a/pkg/configauditreport/controller/resource.go
+++ b/pkg/configauditreport/controller/resource.go
@@ -130,6 +130,7 @@ func (r *ResourceController) SetupWithManager(mgr ctrl.Manager) error {
 		}).For(resource.ForObject, builder.WithPredicates(
 			predicate.Not(predicate.ManagedByTrivyOperator),
 			predicate.Not(predicate.IsBeingTerminated),
+			predicate.Not(predicate.ManagedByKubeEnforcer),
 		)).Owns(resource.OwnsObject).Complete(r.reconcileResource(resource.Kind)); err != nil {
 			return fmt.Errorf("constructing controller for %s: %w", resource.Kind, err)
 		}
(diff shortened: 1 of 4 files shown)


simar7:
Thanks for the PR @mjshastha - do you have a way to measure the memory footprint? I ask this as I'd like to run that too locally and if possible add it to CI testing.

mjshastha:
@simar7 Following the performance improvements seen in KE with the added predicate, we're applying the same fix to the Trivy-operator. We've measured memory usage using Pyroscope and Grafana, and Go benchmarking can also be utilized.

simar7 (approved):

itaysk:
I think it would be good to add an option to skip workloads scanning by a label. now it's kube-bench, tomorrow it's some other infrastructure component, or maybe something the user decides to skip. WDYT? if it makes sense I'll create an issue (low priority, just for tracking)

itaysk:
> does not provide significant additional value over our existing namespace-based exclusions

I think this PR proved that there is additional value, because we needed to create a special predicate just to exclude kube-bench.

> controlled list of system workloads to be excluded

if the discussion is about how to refer to workload to exclude, by name or label, I'm fine with either, or both.

rgarcia89:
I would love to see a possibility to exclude a workload by label. Maybe in the style that we already know from network policies:

```yaml
- namespaceSelector:
    matchLabels:
      kubernetes.io/metadata.name: kube-system
  podSelector:
    matchLabels:
      app: konnectivity-agent
```

I am currently trying to figure out how to exclude gitlab-runner-job pods but not the gitlab-runner itself, which are located in the same namespace.

Issue #2427, opened 2025-02-08
perf: Optimize loading of checks


simar7:
Currently checks are loaded on every workload that gets scanned. This is suboptimal as it make the operator spend most of the time loading checks, even when idle. See the flame graph below for details.

![Image](https://github.com/user-attachments/assets/24e3d1c1-0b48-4167-ac0b-a078788f122d)

This can be optimized by only loading the checks once at initialization.  If there are custom checks that need to get added at a later stage, we can create another compiler and combine the two (built-in and custom) together to accumulate the checks. This would only necessitate the compilation of only the new checks that (if-any) get added and not all checks each time.

simar7:
Upon improving initialization to take place only once the following flame graph was observed. Notice the lack of memory spent on initialization of checks.

![Image](https://github.com/user-attachments/assets/9c064965-25f5-4033-b1d2-803a059e2862)

PoC can be found here: https://github.com/aquasecurity/trivy-operator/tree/pyroscope

PR #2435, opened 2025-02-12
perf: Improve performance by optimizing the loading of checks

simar7:
## Description

This PR aims to improve the performance of Trivy operator.

Overall now trivy operator consumes ~60% less memory footprint while running in a cluster.

### Compared with baseline on the main branch

image

### Top improvements 
image

## Testing methodology 

1. Create deployments with vulnerable images
2. Test harness: https://github.com/simar7/k8s-testing-tools/blob/main/scripts/stress.sh
3. Spin up 50 deployments of vulnerable images, with 1 replica each.
```shell
~/repos/k8s-testing/scripts/stress.sh -r 1
```
4. Run for 10 minutes
5. Teardown
```shell
~/repos/k8s-testing/scripts/stress.sh -d 1
```

## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/2427 

## Checklist

### DIFF
--- a/cmd/trivy-operator/main.go
+++ b/cmd/trivy-operator/main.go
@@ -46,7 +46,7 @@ func main() {
 	runtime.SetMutexProfileFraction(5)
 	runtime.SetBlockProfileRate(5)
 
-	pyroscope.Start(pyroscope.Config{
+	_, _ = pyroscope.Start(pyroscope.Config{
 		ApplicationName: "trivy-operator",
 
 		// replace this with the address of pyroscope server
--- a/cmd/trivy-operator/main.go
+++ b/cmd/trivy-operator/main.go
@@ -3,7 +3,10 @@ package main
 import (
 	"fmt"
 	"os"
+	"runtime"
 
+	"github.com/google/gops/agent"
+	"github.com/grafana/pyroscope-go"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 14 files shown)


cmd/trivy-operator/main.go
```diff
@@ -34,6 +37,44 @@ var (
...
+
+			// these profile types are optional:
+			pyroscope.ProfileGoroutines,
+			pyroscope.ProfileMutexCount,
+			pyroscope.ProfileMutexDuration,
+			pyroscope.ProfileBlockCount,
+			pyroscope.ProfileBlockDuration,
+		},
+	})
+
```
simar7, line 77: note: remove prior to merge

simar7:
@afdesk can you confirm if the test failure seen here is expected? It seems to be that this is the etcd issue isn't it?

afdesk:
> @afdesk can you confirm if the test failure seen here is expected?

yes, you're right! it's the etcd issue. I'll pass this test

afdesk:
I think this workflow is failed with another issue https://github.com/aquasecurity/trivy-operator/actions/runs/13340722929/job/37383562810

there is the log:
```sh
2025/02/18 08:30:57 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
panic: mkdir /home/trivyoperator/.config: read-only file system
goroutine 1 [running]:
main.main()
	/home/runner/work/trivy-operator/trivy-operator/cmd/trivy-operator/main.go:41 +0x3cb

```

simar7:
@afdesk looks like the tests are green now. Could you take another look and approve?

pkg/policy/policy.go
```diff
@@ -109,6 +117,7 @@ func (p *Policies) PoliciesByKind(kind string) (map[string]string, error) {
 	return policies, nil
 }
 
+// TODO: use loaded
```
afdesk, line 120: maybe should we create an issue / discussions for such cases?

afdesk (approved):
If I understood correctly - LGTM
i left one commend - idea, not about the code

pkg/policy/policy.go
```diff
@@ -127,6 +136,31 @@ func (p *Policies) ModulesByKind(kind string) (map[string]string, error) {
...
+	for _, lib := range p.Libraries() {
+		p.loaded = append(p.loaded, lib)
+	}
+
+	for key, policy := range p.data {
+		if !strings.HasSuffix(key, keySuffixRego) {
+			continue
+		}
+		p.loaded = append(p.loaded, policy)
+	}
```
afdesk, line 159: this change doesn't consider `policy.<*>.kinds`.

simar7: It's a good point - but I'd like to understand, when we call a reconciler it in turn calls `Hash()` which will load policies by kind. https://github.com/aquasecurity/trivy-operator/blob/ce7661e3481fac4240389e20606dc3121d71082c/pkg/policy/policy.go#L156 - wouldn't the non-applicable policies be not loaded at that time?

Regardless we should filter by kind here anyway - @nikpivkin do you recall why we did it this way in this PR?

nikpivkin: I think it was done by accident. I think filtering of checks by kind should be added to Policies before evaluation. What about [createPolicyInputFS](https://github.com/aquasecurity/trivy-operator/blob/2fff42f218043488a6f006cd333a64c9e7848ddf/pkg/policy/policy.go#L381) ?

PR #2440, opened 2025-02-13
fix: should work correctly with custom checks

afdesk:
## Related issues
- Close #XXX

## Checklist

### DIFF
--- a/pkg/policy/policy_test.go
+++ b/pkg/policy/policy_test.go
@@ -518,6 +518,50 @@ func TestPolicies_Eval(t *testing.T) {
 			policies:           make(map[string]string),
 			expectedError:      policy.PoliciesNotFoundError,
 		},
+		{
+			name: "Should return empty result for custom rule",
+			resource: &corev1.Pod{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Pod",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "nginx",
+							Image: "nginx:1.16",
+						},
+					},
+				},
+			},
+			useBuiltInPolicies: false,
+			policies: map[string]string{
+				"policy.uses_image_tag_latest.kinds": "Pod",
+				"policy.uses_image_tag_latest.rego": `
+   package trivyoperator.policy.k8s.custom
+
+   import data.lib.result
+   import future.keywords.in
+
+   import data.lib.kubernetes
+
+   __rego_metadata__ := {
+       "id": "recommended labels",
+      "title": "recommended title",
+      "severity": "LOW",
+  }
+
+   failHostNetwork {
+      1 == 0
+   }
+
+   deny[res] {
+        failHostNetwork
+        res := result.new("testing")
+   }`,
+			},
+		},
 	}
 
 	for _, tc := range testCases {


pkg/policy/policy.go
```diff
@@ -276,12 +276,10 @@ func (p *Policies) scannerOptions(policiesFolder string, dataPaths []string, dat
 		rego.WithDataFilesystem(dataFS),
 		rego.WithDataDirs(dataPaths...),
 	}
-	if !hasPolicies && p.cac.GetUseEmbeddedRegoPolicies() {
-		optionsArray = append(optionsArray, rego.WithEmbeddedPolicies(true), rego.WithEmbeddedLibraries(true))
-	} else {
-		optionsArray = append(optionsArray, rego.WithPolicyDirs(policiesFolder))
+	if p.cac.GetUseEmbeddedRegoPolicies() {
```
nikpivkin, line 279: Should embedded checks take precedence over built-in checks? If so, it's worth documenting.

pkg/policy/policy_test.go
```diff
@@ -518,6 +534,116 @@ func TestPolicies_Eval(t *testing.T) {
...
+   failHostNetwork {
+      1 == 0
+   }
+
+   deny[res] {
+        failHostNetwork
+		res := {
+				"msg": "the check should always pass",
+			}
+   }`,
```
nikpivkin, line 562: The rules can be shortened:
```suggestion
   deny[{"msg": "the check should always pass"}]`,
```

pkg/policy/policy_test.go
```diff
@@ -518,6 +534,116 @@ func TestPolicies_Eval(t *testing.T) {
...
+   package trivyoperator.policy.k8s.custom
+   __rego_metadata__ := {
+        "id": "CUSTOMCHECK",
+        "title": "custom check title",
+        "severity": "LOW",
+        "type": "Kubernetes Security Check",
+        "description": "custom check description",
+    }
+
+   failHostNetwork {
```
itaysk, line 593: small thing, why is this rule named as such? there's nothing about host networking in thie check

docs/tutorials/writing-custom-configuration-audit-policies.md
```diff
@@ -30,7 +30,7 @@ __rego_metadata__ := {
 }
 ```
 
-Note that the `recommended_labels` policy is scoped to the `trivyoperator.policy.k8s.custom` package to avoid naming
+Note that the `recommended_labels` policy must be scoped to the `trivyoperator` package to avoid naming
```
itaysk, line 33: maybe also change the package of the example above? I know that it's still valid, but it could be minimized to better  reflect the requirements: `trivyoperator.recommended_labels`

pkg/policy/policy.go
itaysk: I think the behavior should match trivy. in trivy config scan, is it possible to override built-in checks?

pkg/policy/policy_test.go
itaysk: isn't this suggestion negates the check logic? before it has a false statement, now it doesn't

pkg/policy/policy.go
simar7: > I think the behavior should match trivy.

Yes, agreed. Today in Trivy embedded checks are only used as a fallback in case there's something wrong with the bundled checks. As for incorrectly written custom checks, we leave it as the responsibility of the user to provide us with a correct check.


>  in trivy config scan, is it possible to override built-in checks?

Package names must be unique for each check. So technically loading another check with the same package today will result in a compile error for Rego. 

However, you can provide a set of custom checks through a bundle. Trivy always uses the bundle first to load checks. This way you can override any existing checks.

docs/tutorials/writing-custom-configuration-audit-policies.md
afdesk: done https://github.com/aquasecurity/trivy-operator/pull/2440/commits/b9c59969c8d9106a34989ed1f2aa93168172c876

pkg/policy/policy_test.go
```diff
@@ -518,6 +534,107 @@ func TestPolicies_Eval(t *testing.T) {
...
+        "title": "custom check title",
+        "severity": "LOW",
+        "type": "Kubernetes Security Check",
+        "description": "custom check description",
+    }
+
+   deny[{"msg": "this message should be hidden"}]`,
+			},
+			expectedError: "failed to run policy checks on resources",
+		},
```
simar7, line 637: I don't think this testcase is correct because the behavior today is wrong (IMO). I say this because if the check does not apply to a resource, it should not return either a result nor an error. But today we return an error in such a scenario https://github.com/aquasecurity/trivy-operator/blob/b457d371f57ca7ae9a82ce41756c2092f5b6c252/pkg/policy/policy.go#L253-L255

I think we should remove this error return as we should simply return no results and no error. This is possible due to the fact that a check may not apply to a resource.

afdesk: @simar7 Actually, I agree with you, but I'm unsure about my understanding policy checks.

I couldn't reproduce a case when `scanResult` is `nil` for correct policies.
there is a test when I try to apply POD policy to RBAC resource: https://github.com/afdesk/trivy-operator/pull/1/files
`scanResult` contains metadata from the check, but this one isn't passed.

> a check may not apply to a resource.

it seems it's not possible by design.
there is a check before evaluating policies to resources.
https://github.com/afdesk/trivy-operator/blob/9b6b5f463ab2b22e9be214eac43d3a83a7bc6f10/pkg/configauditreport/controller/resource.go#L181-L189

maybe I missed something here

pkg/policy/policy_test.go
afdesk: this place is refactored.

pkg/policy/policy.go
afdesk: IIUC, EmbeddedPolicies and Builtin cannot be applied together, right?
so I thought this package isn't good place to set priority between embedded checks over built-in checks
I'd prefer to check settings here: https://github.com/afdesk/trivy-operator/blob/9b6b5f463ab2b22e9be214eac43d3a83a7bc6f10/pkg/trivyoperator/plugin.go#L87

in any case, I reverted this `if` statement and will fix #2442 in another PR.

wdyt? thanks for your opinion and time

simar7: > IIUC, EmbeddedPolicies and Builtin cannot be applied together, right? so I thought this package isn't good place to set priority between embedded checks over built-in checks I'd prefer to check settings here: https://github.com/afdesk/trivy-operator/blob/9b6b5f463ab2b22e9be214eac43d3a83a7bc6f10/pkg/trivyoperator/plugin.go#L87
> 
> in any case, I reverted this `if` statement and will fix #2442 in another PR.
> 
> wdyt? thanks for your opinion and time

Sure makes sense to me.

pkg/policy/policy_test.go
simar7: As we discussed offline - the only cases where scanResult & err both can be nil are if: 1. setup of checks fails, 2. namespace mismatch. 3. kind mismatch

Since we know 3. cannot happen at this stage (subtype filtering of kind is already performed) and 1. isn't the case at this stage, that would leave us with only 2. as the potential reason.

After testing this further, I could confirm that it is indeed the case as this logic is the one that takes place https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/scanner.go#L194-L197

simar7 (approved):

pkg/policy/policy_test.go
```diff
@@ -529,11 +646,11 @@ func TestPolicies_Eval(t *testing.T) {
...
 			if tc.expectedError != "" {
 				g.Expect(err).To(HaveOccurred())
-				// FIXME: Assert result
-			} else {
-				g.Expect(err).ToNot(HaveOccurred())
-				g.Expect(reflect.DeepEqual(getPolicyResults(checks), tc.results))
+				return
 			}
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(tc.results).Should(ContainElements(getPolicyResults(checks)))
```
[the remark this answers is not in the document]
simar7: I updated it here https://github.com/aquasecurity/trivy-operator/pull/2440/commits/2fca81f6a3946754d877ba4a468258f0dd78f461 - was that what you were looking for @nikpivkin?

simar7:
@itaysk and @nikpivkin let us know if you still have any more comments as we'd like to include this change in the next release.

pkg/policy/policy_test.go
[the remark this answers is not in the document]
afdesk: @simar7 I think @nikpivkin meant that `tc.results` should be replaced `checks`.
let me fix it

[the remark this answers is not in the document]
afdesk: @simar7 actually, @nikpivkin arised an interesting case - this test is failed
https://github.com/aquasecurity/trivy-operator/blob/5a93cbe94ae71979e2009b90f7967b14d4394baf/pkg/policy/policy_test.go#L485-L501

but it passed for `g.Expect(reflect.DeepEqual(getPolicyResults(checks), tc.results))`.

`len(checks) == 1` and `len(tc.results) == 145`, so  `checks` and `tc.results` aren't equal

screenshot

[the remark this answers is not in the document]
afdesk: just FYI, the same behavior for `main` branch
screenshot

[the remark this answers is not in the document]
nikpivkin: The original assertion did not compare the result to anything, so the test succeeded. It should have been `g.Expect(reflect.DeepEqual(getPolicyResults(checks), tc.results)).Should(BeTrue())`.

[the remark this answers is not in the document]
nikpivkin: Maybe we should revert this change, add a TODO and come back to it later, since it's not an issue with this PR?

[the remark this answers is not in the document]
simar7: Yeah sounds good to me.

> Maybe we should revert this change, add a TODO and come back to it later, since it's not an issue with this PR?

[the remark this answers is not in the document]
afdesk: doing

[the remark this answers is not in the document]
afdesk: https://github.com/aquasecurity/trivy-operator/pull/2440/commits/9b90f24edf8c0535fb405a8f161412d812e2d04e

nikpivkin (approved):

itaysk:
> @itaysk and @nikpivkin let us know if you still have any more comments as we'd like to include this change in the next release.

don't let me block this

pkg/policy/policy_test.go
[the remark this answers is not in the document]
afdesk: created #2458

PR #2507, opened 2025-04-03
chore(deps): Bump `trivy-*` deps

simar7:
## Description

chore(deps): Bump trivy-* deps


pkg/policy/loader.go
```diff
@@ -116,7 +116,10 @@ func (pl *policyLoader) getBuiltInPolicies(ctx context.Context) ([]string, error
 	if err = client.DownloadBuiltinChecks(ctx, pl.RegistryOptions); err != nil {
 		return nil, xerrors.Errorf("failed to download built-in policies: %w", err)
 	}
-	return client.LoadBuiltinChecks()
+	var paths []string
+	paths = append(paths, client.LoadBuiltinChecks())
+
+	return paths, nil
```
afdesk, line 122:
```suggestion
	return []string{ client.LoadBuiltinChecks() }, nil
```

afdesk: This change doesn't matter, but it's more readable for me

afdesk (approved):

pkg/policy/loader.go
simar7: oh nice - yeah I will update that.

PR #2529, opened 2025-04-28
feat: add using Trivy config files

### LINKED ISSUE #2274: Allow to try several mirrors for a given registry
We use different mirrors and this works correctly with [containerd](https://github.com/containerd/containerd/blob/main/docs/hosts.md#host-fields-in-the-toml-table-format) which tries them in order and use the first working one (i.e. not 404 or other errors).

We need the same feature for trivy-operator.

This may require feature addition in trivy itself.

afdesk:
## Description
This PR adds support of [Trivy config file](https://trivy.dev/latest/docs/references/configuration/config-file/#registry-options). it needs because Trivy allows set up mirrors for remote container and databases only via Trivy config file.

```sh
config="registry:
  mirrors:
    index.docker.io:
     - mirror.gcr.io"

kubectl patch cm trivy-operator-trivy-config -n trivy-system \
  --type merge \
  -p "$(jq -n --arg config "$config" '{"data": {"trivy.configFile": $config}}')"

```

## Related issues
- Close #2274 

## Checklist

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -129,6 +129,7 @@ Keeps security report resources updated
 | trivy.additionalVulnerabilityReportFields | string | `""` | additionalVulnerabilityReportFields is a comma separated list of additional fields which can be added to the VulnerabilityReport. Supported parameters: Description, Links, CVSS, Target, Class, PackagePath and PackageType |
 | trivy.clientServerSkipUpdate | bool | `false` | clientServerSkipUpdate is the flag to enable skip databases update for Trivy client. Only applicable in ClientServer mode. |
 | trivy.command | string | `"image"` | command. One of `image`, `filesystem` or `rootfs` scanning, depending on the target type required for the scan. For 'filesystem' and `rootfs` scanning, ensure that the `trivyOperator.scanJobPodTemplateContainerSecurityContext` is configured to run as the root user (runAsUser = 0). |
+| trivy.configFile | string | `nil` | configFile can be used to tell Trivy to use specific options available only in the config file (ex. Mirror registries). |
 | trivy.createConfig | bool | `true` | createConfig indicates whether to create config objects |
 | trivy.dbRegistry | string | `"mirror.gcr.io"` |  |
 | trivy.dbRepository | string | `"aquasec/trivy-db"` |  |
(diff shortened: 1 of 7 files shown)


deploy/helm/README.md
```diff
@@ -129,6 +129,7 @@ Keeps security report resources updated
 | trivy.additionalVulnerabilityReportFields | string | `""` | additionalVulnerabilityReportFields is a comma separated list of additional fields which can be added to the VulnerabilityReport. Supported parameters: Description, Links, CVSS, Target, Class, PackagePath and PackageType |
 | trivy.clientServerSkipUpdate | bool | `false` | clientServerSkipUpdate is the flag to enable skip databases update for Trivy client. Only applicable in ClientServer mode. |
 | trivy.command | string | `"image"` | command. One of `image`, `filesystem` or `rootfs` scanning, depending on the target type required for the scan. For 'filesystem' and `rootfs` scanning, ensure that the `trivyOperator.scanJobPodTemplateContainerSecurityContext` is configured to run as the root user (runAsUser = 0). |
+| trivy.configFile | string | `nil` | configFile can be used to tell Trivy to use specific options available only in the config file (ex. Mirror registries). |
```
simar7, line 132:
```suggestion
| trivy.configFile | string | `nil` | configFile can be used to tell Trivy to use specific options available only in the config file (e.g. Mirror registries). |
```

deploy/helm/values.yaml
```diff
@@ -453,6 +453,15 @@ trivy:
   #   - CVE-1970-0001
   #   - CVE-1970-0002
 
+  # -- configFile can be used to tell Trivy to use specific options available only in the config file (ex. Mirror registries).
```
simar7, line 456:
```suggestion
  # -- configFile can be used to tell Trivy to use specific options available only in the config file (e.g. Mirror registries).
```

deploy/helm/values.yaml
```diff
@@ -453,6 +453,15 @@ trivy:
...
+  configFile: ~
```
simar7, line 457: does the `~` imply anything?

afdesk: The tilde is one of the ways the null value can be written. 
I thought it's a common style for this value.yaml:
https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/values.yaml#L187-L188

simar7 (approved):

bh-tt:
This PR failed to include the GetPodSpecForClientServerMode in the change, resulting in failing jobs when using this feature together with the trivy server.

PR #2555, opened 2025-05-14
chore: replace deprecated '--slow' with '--parallel'

henkka:
Relates to https://github.com/aquasecurity/trivy/pull/5572 where --slow was deprecated in the `trivy` project.

With current implementation, each scan job logs the following warning:

```
WARN    "--slow" is deprecated. Use "--parallel 1" instead.
```

This pull request modifies `trivy-operator` to use the `--parallel 1` instead of the deprecated `--slow` flag, while keeping the `trivy-operator`'s own API backward compatible (ie. call it `slow` but under the hood convert it to `parallel`)

### DIFF
--- a/pkg/plugins/trivy/flags.go
+++ b/pkg/plugins/trivy/flags.go
@@ -29,7 +29,7 @@ func Slow(c Config) string {
 		return ""
 	}
 	if c.GetSlow() {
-		return "--slow"
+		return "--parallel 1"
 	}
 	return ""
 }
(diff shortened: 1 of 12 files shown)


pkg/plugins/trivy/flags.go
```diff
@@ -29,7 +29,7 @@ func Slow(c Config) string {
 		return ""
 	}
 	if c.GetSlow() {
-		return "--slow"
+		return "--parallel 1"
```
afdesk, line 32: maybe we should support backward compatibility with older tags too
`--slow` was deprecated in v0.53.0 https://github.com/aquasecurity/trivy/discussions/7061

afdesk (changes requested):

afdesk:
@henkka thanks for your contribution
i left one small comment

pkg/plugins/trivy/flags.go
henkka: Thank you for the quick turnaround! Fixed with https://github.com/aquasecurity/trivy-operator/pull/2555/commits/da40ff16c22ee572938a8d7ffc700474f35aff6d

afdesk (approved):

PR #2582, opened 2025-05-31
fix: enable more go-critic rules

mmorel-35:
## Description

This enables the following go-critic rules:

* [indexAlloc](https://go-critic.com/overview.html#indexalloc)
* [octalLiteral](https://go-critic.com/overview.html#octalliteral)
* [regexpSimplify](https://go-critic.com/overview.html#regexpsimplify)
* [sloppyReassign](https://go-critic.com/overview.html#sloppyreassign)

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,7 +49,6 @@ linters:
         - commentedOutCode
         - hugeParam
         - rangeValCopy
-        - regexpSimplify
         - sloppyReassign
         - unnamedResult
         - whyNoLint
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,7 +48,6 @@ linters:
         - appendAssign
         - commentedOutCode
         - hugeParam
-        - indexAlloc
         - octalLiteral
         - rangeValCopy
         - regexpSimplify
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,7 +48,6 @@ linters:
         - appendAssign
         - commentedOutCode
         - hugeParam
-        - octalLiteral
         - rangeValCopy
         - regexpSimplify
         - sloppyReassign
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,7 +49,6 @@ linters:
         - commentedOutCode
         - hugeParam
         - rangeValCopy
-        - sloppyReassign
         - unnamedResult
         - whyNoLint
       enabled-tags:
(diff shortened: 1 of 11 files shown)


pkg/apis/aquasecurity/v1alpha1/compliance_types.go
```diff
@@ -170,7 +170,8 @@ type ComplianceCheck struct {
 // ToComplianceSpec map data from crd compliance spec to trivy compliance spec
 func ToComplianceSpec(cSpec Compliance) spec.ComplianceSpec {
 	specControls := make([]defsecTypes.Control, 0)
-	for _, control := range cSpec.Controls {
+	for i := range cSpec.Controls {
+		control := &cSpec.Controls[i]
```
simar7: for my understanding, which rule is this and why is it helpful?

mmorel-35: This is [rangeValCopy](https://go-critic.com/overview.html#rangevalcopy) it is using a pointer to the accessed resource instead of the direct resource

simar7: Thanks - while I agree it could be useful for big objects, it seems a bit premature to optimize on every single call to me as it affects readability.

mmorel-35: There is an option in https://golangci-lint.run/usage/linters/#gocritic to determine the size in bytes that makes the warning trigger, the default is 128 bytes.

simar7: Personally as it's a bit tedious to implement each time, I would remove it for now unless we can determine what kind of performance impact this change would make. @afdesk wdyt?

mmorel-35: I found the go-critic quite clear on how to achieve it but that wasn't obvious before knowing it

simar7: Personally it feels a bit clunky to me to write this way. Can we take this out for now and maybe if it does really impact performance (we need to measure that), implement it in a separate PR?

mmorel-35: Of course ! I'll remove this part.

simar7 (approved):

PR #2585, opened 2025-06-03
ci: bump up KinD to the latest version - v0.29.0

afdesk:
## Description
This PR bumps up KinD to the latest version and also fixes a warning about using `@sha256`:
```
Please include the @sha256: image digest from the image in the release notes. You can find available image tags on the release page, https://github.com/kubernetes-sigs/kind/releases/tag/v0.24.0

```

## Checklist

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -184,8 +184,8 @@ jobs:
           args: release -f=goreleaser-e2e.yaml --snapshot --skip=publish --clean
       - name: Install kind and create cluster
         run: >
-          curl -Lo ./kind https://kind.sigs.k8s.io/dl/${{ env.KIND_VERSION
-          }}/kind-linux-amd64
+          # Using outdated version of kind to test cluster vulnerability reports
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.17.0/kind-linux-amd64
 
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yaml
```diff
@@ -184,8 +184,9 @@ jobs:
           args: release -f=goreleaser-e2e.yaml --snapshot --skip=publish --clean
       - name: Install kind and create cluster
         run: >
-          curl -Lo ./kind https://kind.sigs.k8s.io/dl/${{ env.KIND_VERSION
-          }}/kind-linux-amd64
+          # Using outdated version of kind to test cluster vulnerability reports
+
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.24.0/kind-linux-amd64
```
simar7: So we will have two kinds at different versions? What stops us from using v0.29.0 for everything?

afdesk: I think it was intended to use an old KinD with an old cluster version to test the cluster vuln reports.

but actually, you're right, we can use the same version of KinD, but with another node version.

done https://github.com/aquasecurity/trivy-operator/pull/2585/commits/31a62be61b78fc27c2a2a12cc3a4239e790d4396

simar7 (approved):

PR #2595, opened 2025-06-06
Fix/login private registry

afdesk:
## Related issues
- Close #XXX

## Checklist

### DIFF
--- a/pkg/plugins/trivy/image.go
+++ b/pkg/plugins/trivy/image.go
@@ -31,6 +31,15 @@ func (j *ImageJobSpecMgr) GetPodSpec(ctx trivyoperator.PluginContext, config Con
 	return j.getPodSpecFunc(ctx, config, workload, credentials, securityContext, p, clusterSboms)
 }
 
+// GetPodSpec creates a PodSpec for the Trivy image scan job.
+//
+// The number of main containers correspond to the number of containers
+// defined for the scanned workload. Each container runs the Trivy image scan
+// command and skips the database download.
+//
+//	trivy --cache-dir /tmp/trivy/.cache image --skip-update \
+//	  --format json <container image>
+//
 // In the Standalone mode there is the init container responsible for
 // downloading the latest Trivy DB file from GitHub and storing it to the
 // emptyDir volume shared with main containers. In other words, the init
@@ -38,13 +47,16 @@ func (j *ImageJobSpecMgr) GetPodSpec(ctx trivyoperator.PluginContext, config Con
 //
 //	trivy --cache-dir /tmp/trivy/.cache image --download-db-only
 //
-// The number of main containers correspond to the number of containers
-// defined for the scanned workload. Each container runs the Trivy image scan
-// command and skips the database download:
+// In the ClientServer each container runs Trivy image scan command and refers to Trivy server URL
+// returned by Config.GetServerURL:
 //
-//	trivy --cache-dir /tmp/trivy/.cache image --skip-update \
-//	  --format json <container image>
-func GetPodSpecForStandaloneMode(ctx trivyoperator.PluginContext,
+//		trivy image --server <server URL> \
+//		  --format json <container image>
+//
+//	 Also there is the init container responsible for downloading the latest Trivy Java DB for both modes
+//
+//		trivy --cache-dir /tmp/trivy/.cache image --download-java-db-only
+func GetPodSpecForImageScan(ctx trivyoperator.PluginContext,
 	config Config,
 	workload client.Object,
 	credentials map[string]docker.Auth,
@@ -79,24 +91,7 @@ func GetPodSpecForStandaloneMode(ctx trivyoperator.PluginContext,
 		secrets = append(secrets, secret)
 	}
 
-	trivyImageRef, err := config.GetImageRef()
-	if err != nil {
-		return corev1.PodSpec{}, nil, err
-	}
-
 	trivyConfigName := trivyoperator.GetPluginConfigMapName(Plugin)
-
-	dbRepository, err := config.GetDBRepository()
-	if err != nil {
-		return corev1.PodSpec{}, nil, err
-	}
-
-	requirements, err := config.GetResourceRequirements()
-	if err != nil {
-		return corev1.PodSpec{}, nil, err
-	}
-
-	cacheDir := config.GetImageScanCacheDir()
 
 	volumeMounts := []corev1.VolumeMount{
 		{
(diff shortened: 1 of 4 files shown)


simar7:
> NOTE:
By design, the operator does not allow setting different usernames and passwords for multiple registries. Therefore, when downloading trivy-db and trivy-java-db from repositories that require authentication, you must use the same registry for both. Authentication will use the registry specified in the trivy.dbRepository field.

Is this documented somewhere? If not, we should as it's a limitation today which may impact users.

afdesk:
> Is this documented somewhere? If not, we should as it's a limitation today which may impact users.

sure, I've added the next block here https://aquasecurity.github.io/trivy-operator/latest/docs/vulnerability-scanning/private-registries/

pkg/plugins/trivy/image.go
```diff
@@ -115,52 +110,62 @@ func GetPodSpecForStandaloneMode(ctx trivyoperator.PluginContext,
...
+	if err != nil {
+		return corev1.PodSpec{}, nil, err
+	}
+
+	if config.TrivyDBRepositoryCredentialsSet() {
+		dockerConfigVolumeMount := []corev1.VolumeMount{
+			{
+				Name:      dockerConfigVolumeName,
+				ReadOnly:  false,
+				MountPath: "/root/.docker",
```
simar7: Just for my own understanding, will this mount point be the same across different environments? Should we dynamically determine this?

afdesk: you're right in common case - this dir can be different for different systems.
but in our case - we run a pod without security context, and it runs in the same docker image (trivy) based on alpine, so "/root/.docker" should be the same.

afdesk: now it uses a specific docker folder instead of a default dir
https://github.com/aquasecurity/trivy-operator/pull/2595/commits/7c6576689feb6ec5ec07b202bcd074b8c222070f

afdesk:
@simar7 I finally re-tested this PR and it's ready for your review.

afdesk:
@simar7 rebased the main branch and tested on Minikube

pkg/plugins/trivy/plugin_test.go
```diff
@@ -2201,110 +2201,110 @@ default ignore = false`,
...
+								Name: "TRIVY_IGNORE_UNFIXED",
 								ValueFrom: &corev1.EnvVarSource{
 									ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
 										LocalObjectReference: corev1.LocalObjectReference{
 											Name: "trivy-operator-trivy-config",
 										},
-										Key:      "trivy.httpsProxy",
+										Key:      "trivy.ignoreUnfixed",
 										Optional: ptr.To[bool](true),
 									},
```
simar7, line 2224: why did the ordering of these test cases change? are you using a different linter?

pkg/plugins/trivy/plugin_test.go
```diff
@@ -7697,6 +7717,12 @@ func TestGetInitContainers(t *testing.T) {
 				},
 			).Build()
 
+			// a number of init containers is 2 for Standalone, and 1 for ClientServer mode
+			numInitContainers := 2
+			if tc.configData["trivy.mode"] == string(trivy.ClientServer) {
+				numInitContainers = 1
+			}
+
```
simar7: I think this test has way too many cases. Can we split it up based on client server vs standalone mode?

pkg/plugins/trivy/jobspec.go
```diff
@@ -51,22 +51,6 @@ type PodSpecMgr interface {
 func NewPodSpecMgr(config Config) PodSpecMgr {
```
simar7, line 51: Can we add a TODO to the callers of `NewPodSpecMgr()` to check if `err != nil`? Currently we do not and this will result in a panic here: https://github.com/aquasecurity/trivy-operator/blob/1c0f3d938fb16b3ab67c1599b2dccebde46a51bb/pkg/plugins/trivy/plugin.go#L110

pkg/plugins/trivy/plugin_test.go
afdesk: previously `GetPodSpecForStandaloneMode` and `GetPodSpecForClientServerMode` had a lot of duplicated code, especially around creating env. the order of environment variable creation was a bit different

pkg/plugins/trivy/jobspec.go
afdesk: It seems I missed why this will result in a panic. IIUC, `podSpec` isn't a pointer and is always defined.

pkg/plugins/trivy/plugin_test.go
afdesk: i tried - https://github.com/aquasecurity/trivy-operator/pull/2595/commits/7cb23fbb4796540cad642f185ba687837e3042c6
but it looks not good for me (

afdesk:
@simar7 could you pls take another look? thanks

pkg/plugins/trivy/plugin_test.go
afdesk: actually, it's splitted here now https://github.com/aquasecurity/trivy-operator/pull/2595/commits/7cb23fbb4796540cad642f185ba687837e3042c6

pkg/plugins/trivy/jobspec.go
afdesk: @simar7 I thought a lot, but it seems have to disturb you.

Just to clarify — `NewPodSpecMgr()` currently never returns nil and doesn’t have any error conditions (it always returns a valid `PodSpecMgr`).

Could you please confirm where a panic could actually occur here?

afdesk:
@simar7 could you please prioritize this PR?

The key change here is the removal of duplicate logic shared between the server and standalone modes.

This will simplify maintenance, reduce the risk of inconsistencies, and make future changes much easier to implement.

simar7:
> @simar7 could you please prioritize this PR?
> 
> 
> 
> The key change here is the removal of duplicate logic shared between the server and standalone modes.
> 
> 
> 
> This will simplify maintenance, reduce the risk of inconsistencies, and make future changes much easier to implement.

I'll take a look at once I'm back. So we could aim it for the November end release.

afdesk:
> I'll take a look at once I'm back. So we could aim it for the November end release.

@simar7 sure! have a good vacation!

pkg/plugins/trivy/jobspec.go
simar7: If there's an unsupported workload, this would return an empty structure https://github.com/aquasecurity/trivy-operator/blob/main/pkg/kube/object.go#L264

In such a case, should we check the error first prior to using podSpec?

afdesk: i realized your concern about this place, and tried to improve readability here: https://github.com/aquasecurity/trivy-operator/pull/2595/commits/d61ec7462dde5a9e4e13c731c54337a5c2db593e

simar7:
@afdesk could you rebase this PR?

afdesk:
> @afdesk could you rebase this PR?

done

PR #2612, opened 2025-06-13
perf: Optimized ConfigMap Caching: Storing Metadata Only.

mjshastha:
As part of this implementation, we are implementing the following changes aimed at improving memory consumption:

Currently, ConfigMaps are stored in the cache including data. - In the new approach will store only the metadata of the ConfigMap. This will reduce the memory footprint by including only the essential metadata.


simar7:
@mjshastha could you fix the tests?

simar7:
@mjshastha have you had a chance to run this on a perf test to see if this actually solves the performance issue that you're facing?

Issue #2629, opened 2025-06-27
ConfigMap predicate excludes required resources, causing policy load failures.


mjshastha:
The current predicate logic is excluding the TrivyConfigMapName and PoliciesConfigMapName, while our intent is to watch only these two.
https://github.com/aquasecurity/trivy-operator/blob/main/pkg/configauditreport/controller/checks.go#L104

We’re observing numerous frequent errors in the Trivy Operator logs like:

[...]

PR #2631, opened 2025-06-28
fix: correct predicate logic to allow whitelisted ConfigMaps

nikpivkin:
## Related issues
- Close https://github.com/aquasecurity/trivy-operator/issues/2629

## Checklist

### DIFF
--- a/pkg/configauditreport/controller/checks.go
+++ b/pkg/configauditreport/controller/checks.go
@@ -102,7 +102,7 @@ var allowedConfigMaps = set.New(
 var configPredicate = func(namespace string) predicate.Predicate {
 	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
 		if allowedConfigMaps.Contains(obj.GetName()) {
-			return false
+			return true
 		}
 		return obj.GetNamespace() == namespace
 	})


nikpivkin:
@simar7 Does installation namespace refer to the operator's namespace?

https://github.com/aquasecurity/trivy-operator/blob/0abe0d505c31cbc39bd3857c4d5cd7ce43fd3355/docs/docs/configuration-auditing/built-in-policies.md?plain=1#L3-L5

If so, the current Reconciler logic is incorrect — it watches ConfigMaps in all namespaces.

nikpivkin:
@simar7 During integration tests, I noticed that resources from the `trivy-system` namespace are being scanned, even though I specified `default` as the target namespaces. I logged the `InstallModePredicate` predicate and its output ok, but the resource is still being filtered:

```bash
2025-06-30T21:00:42+06:00     INFO    predicate       install mode predicate  {"mode": "SingleNamespace", "operatorNamespace": "trivy-system", "targetNamespaces": ["default"], "allow": false, "objNamespace": "trivy-system"}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "Service", "name": {"name":"trivy-operator","namespace":"trivy-system"}}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Configuration audit report exists       {"kind": "Service", "name": {"name":"trivy-operator","namespace":"trivy-system"}}
```

Anything I missed? 

I also encountered that the scan goes into a loop for some unknown reason and I see repeated logs:
```bash
2025-06-30T21:00:42+06:00     INFO    predicate       install mode predicate  {"mode": "SingleNamespace", "operatorNamespace": "trivy-system", "targetNamespaces": ["default"], "allow": false, "objNamespace": "trivy-system"}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "Service", "name": {"name":"trivy-operator","namespace":"trivy-system"}}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Configuration audit report exists       {"kind": "Service", "name": {"name":"trivy-operator","namespace":"trivy-system"}}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
2025-06-30T21:00:42+06:00     DEBUG   resourcecontroller      Configuration audit report exists       {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
```

By the way, should the `aggregate-config-audit-reports-view` (and similar) resource be excluded if I don't want to scan the `trivy-system` and have configured this? Right now predicates don't filter out such resources, as they are marked as not managed by Trivy (`app.kubernetes.io/managed-by: kubectl`) and I don't see them in the predicate logs, only in the resource controller:
```bash
  2025-06-30T21:16:59+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
  2025-06-30T21:16:59+06:00     DEBUG   resourcecontroller      Checking whether configuration audit report exists      {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
  2025-06-30T21:16:59+06:00     DEBUG   resourcecontroller      Configuration audit report exists       {"kind": "ClusterRole", "name": {"name":"aggregate-config-audit-reports-view"}}
```

nikpivkin:
@simar7 The `ChecksLoader` and `PolicyConfigController` are currently working inconsistently. Both react to changes in the ConfigMap: `ChecksLoader` loads the policies, while `PolicyConfigController` deletes outdated reports based on the policy hash for each kind. A synchronization mechanism between these controllers is needed. Two possible approaches:

1. Move the policy loading logic into `PolicyConfigController`, adjust the ConfigMap watch conditions, and perform report cleanup in a loop for all required kinds—without creating a separate controller for each.

2. Introduce a CRD, e.g., `ConfigScanRequest`, which is created after policies are loaded and contains the computed hashes for all relevant kinds. `PolicyConfigController` would then watch this resource and trigger report cleanup accordingly.

It would also be nice to add tests for this.

I don't have much work with k8s operators, so I may be missing something. WDYT?

nikpivkin:
> I also encountered that the scan goes into a loop for some unknown reason and I see repeated logs:

I think the logs were so noisy because `PolicyConfigController ` creates a controller for a large number of kind and I didn't clear all resources completely and scale the trivy-operator to zero replicas before running the tests. I'll leave the script for clearing resources here:
```bash
#!/usr/bin/env bash
kubectl scale deployment trivy-operator -n trivy-system --replicas=0
kubectl delete jobs -n trivy-system --all
kubectl delete vuln -n trivy-system --all
# kubectl delete clustervuln -n trivy-system --all
# kubectl delete sboms --all
kubectl delete exposedsecret -n trivy-system --all
kubectl delete configaudit -n trivy-system --all
# kubectl delete clusterconfigaudit --all
kubectl delete rbacassessment -n trivy-system --all
kubectl delete infraassessment -n trivy-system --all
# kubectl delete clusterinfraassessment --all
# kubectl delete clusterrbacassessmentreport --all
kubectl delete scanreq -n trivy-system --all
```



> 2\. Introduce a CRD, e.g., `ConfigScanRequest`, which is created after policies are loaded and contains the computed hashes for all relevant kinds. `PolicyConfigController` would then watch this resource and trigger report cleanup accordingly.

I implemented this item and everything works fine, now ChecksLoader and PolicyConfigController work in a consistent way. I am interested in your opinion. Adding a new CRD should not affect the end users, right?

The e2e tests failed, but the logs showed nothing due to the wrong way of specifying commands. I opened a [PR](https://github.com/aquasecurity/trivy-operator/pull/2637) with the fix for the action.

simar7:
> @simar7 Does installation namespace refer to the operator's namespace?
> 
> https://github.com/aquasecurity/trivy-operator/blob/0abe0d505c31cbc39bd3857c4d5cd7ce43fd3355/docs/docs/configuration-auditing/built-in-policies.md?plain=1#L3-L5
> 
> If so, the current Reconciler logic is incorrect — it watches ConfigMaps in all namespaces.

Yes your interpretation is correct, where's the logic which makes it watch all configmaps today?

simar7:
> Move the policy loading logic into PolicyConfigController, adjust the ConfigMap watch conditions, and perform report cleanup in a loop for all required kinds—without creating a separate controller for each.


I think this approach is simpler. Creating another CRD also puts more pressure on the existing nodes. Plus we don't really needs another resource if there is only one controller (no sync needed). What are the downsides of such an approach?

nikpivkin:
> Yes your interpretation is correct, where's the logic which makes it watch all configmaps today?

This will happen if we just invert the logic rather than fixing the predicate.
https://github.com/aquasecurity/trivy-operator/blob/f55b1fcad791ef9fc7252b49619ab763765ef88f/pkg/configauditreport/controller/checks.go#L104-L106
I fixed the predicate in this PR.

> What are the downsides of such an approach?

I will investigate this and create a separate branch from this PR for comparison.

nikpivkin:
> I think this approach is simpler.

I've taken a closer look at the `PolicyConfigController`, and I now believe this approach isn't feasible. The `PolicyConfigController` creates controllers for each required resource that uses requeuing for batch deletion. If we switch to a single controller. If we switch to a single controller:

- Requeuing will no longer be possible
- Using one controller for all resources can affect performance

simar7:
@nikpivkin should we in that case first revert the checks loader logic completely as today it's not adding any performance improvements (but introducing a new problem) - after this we could try out sorting the other issue.

nikpivkin:
> it's not adding any performance improvements

Why? Didn't we run benchmarks and it improved performance significantly?

What about just fixing the predicate logic without adding tests, but adding them later when we come to a solution?

simar7:
> What about just fixing the predicate logic without adding tests, but adding them later when we come to a solution?

I think this might be the way to go - we do understand why the testing will fail so it isn't a cause that we're overlooking.

simar7 (approved):

PR #2632, opened 2025-07-01
fix: Deletion of scanjob before ttl expires

### LINKED ISSUE #2362: Vulnerability ScanJob is immediately deleted

tom1299:
## Description
Currently jobs are directly deleted after they are completed by the trivy-operator without honouring the jobs `ttlSecondsAfterFinished` and thus also the helm configuration parameter `operator.scanJobTTL`.
The change introduced in this PR will only delete the job if it has no `ttlSecondsAfterFinished`.

**Note** that this PR has the same content as https://github.com/aquasecurity/trivy-operator/pull/2375 with the changes done on the related branch based on the newest main branch.

## Related issues
- Close #2362 

## Checklist
- [x ] I've added tests that prove my fix is effective or that my feature works.


tom1299:
@simar7 Sorry for the slight delay but it took me some time to get the integration tests to work. I did two additional changes compared to #2375:
* Replace the new image replacement in the [kustomization.yaml](https://github.com/aquasecurity/trivy-operator/pull/2632/files#diff-9cb2a73189269f35d2f7af6c6b5c3536ce391e3b4682eaeafe9d3efc03d0fcf4) with a patch which deletes the deployment and added a comment. One the one hand this explains why the deployment should not be deployed and on the other hand the deployment is actually not deployed. While with the former image replacement, the pod of the deployment was just hanging in CrashLoopBackoff.
* Added code for changing the default configuration as part of the integration testes [here](https://github.com/aquasecurity/trivy-operator/pull/2632/files#diff-842eaa5de0f3e4f62427f86b028435c268fbc7a56a85085038c5fbe803d2eb8a) since the integration tests will not work with the default ttl for scan jobs of "0".

Please review these changes. Any feedback is welcome 🙂

afdesk (approved):

afdesk:
@tom1299 thanks for your contribution! 
LGTM!

simar7 (approved):

PR #2636, opened 2025-07-02
fix: Fix Helm global.image.registry and overrides

marcofranssen:
## Description

Previously it was impossible to use the `global.image.registry` because that would also override the nodeCollector image.

In this PR I have harmonized the interface to put all image related values in a image object.

Furthermore I fixed the image rendering to always take the specific image values first and fall back to the global.image.registry if an image doesn't provide a value for registry.

See here for the working solution:

```shell
$ helm template --set global.image.registry=docker.io ./deploy/helm | grep -E 'image:|repository|aquasec/|aquasecurity/'
  node.collector.imageRef: "ghcr.io/aquasecurity/node-collector:0.3.1"
  policies.bundle.oci.ref: "docker.io/aquasec/trivy-checks:1"
  trivy.repository: "docker.io/aquasec/trivy"
  trivy.dbRepository: "docker.io/aquasec/trivy-db"
  trivy.javaDbRepository: "docker.io/aquasec/trivy-java-db"
          image: "docker.io/aquasec/trivy-operator:0.27.1
```

Without this fix all images would be docker.io and therefore breaking the nodeCollector image.

See here the result of the updated Helm documentation https://github.com/aquasecurity/trivy-operator/blob/87e066436668f4809fe46d9f9bc53cc548158878/deploy/helm/README.md

## Checklist

### DIFF
--- a/deploy/helm/templates/configmaps/trivy.yaml
+++ b/deploy/helm/templates/configmaps/trivy.yaml
@@ -86,7 +86,7 @@ data:
   {{- end }}
   {{- if .Values.trivy.ignoreFile }}
   trivy.ignoreFile: |
-{{- range .Values.trivy.ignoreFile }}
+  {{- range .Values.trivy.ignoreFile }}
     {{ . | indent 4 }}
   {{- end }}
   {{- end }}
@@ -131,8 +131,8 @@ data:
   TRIVY_CACHE_DIR: "/home/scanner/.cache/trivy"
   TRIVY_DEBUG: {{ .Values.trivy.debug | quote }}
   TRIVY_SKIP_DB_UPDATE: "false"
-  TRIVY_DB_REPOSITORY: "{{ .Values.trivy.dbRegistry }}/{{ .Values.trivy.dbRepository }}"
-  TRIVY_JAVA_DB_REPOSITORY: "{{ .Values.trivy.javaDbRegistry }}/{{ .Values.trivy.javaDbRepository }}"
+  TRIVY_DB_REPOSITORY: {{ include "trivy-operator.imageRepository" (tuple .Values.trivy.db.image .Values.global.image ) | quote }}
+  TRIVY_JAVA_DB_REPOSITORY: {{ include "trivy-operator.imageRepository" (tuple .Values.trivy.javaDb.image .Values.global.image ) | quote }}
   TRIVY_INSECURE: "{{ .Values.operator.builtInServerRegistryInsecure }}"
   {{- end }}
 {{- end }}


marcofranssen:
@afdesk @nikpivkin @simar7 Could someone have a look at this fix and cut a release?

I would love to fix our deployment that uses the global.image.

Or maybe at the minimum trigger the workflows, so I can see if there is anything else I ened to fix to comply with the CI pipelines.

marcofranssen:
@afdesk I have updated the README.md to include the newline at the end.

marcofranssen:
@afdesk I have rebased the branch on latest main. Any timeline when the changes in this PR can be released?

afdesk:
@marcofranssen thanks a lot for your contribution! we really appreciate it

`e2e-test` is failed twice, do you understand a reason?

also, it looks like `BREAKING CHANGE`, so we need to think it over.

marcofranssen:
@afdesk It is indeed a breaking change when it comes to value structure. So might require at least a minor version bump for the Chart.

However it also fixes the image values to be usable in a way most of the Charts out there allow to override the global setting on a component basis. So it also standardizes on how end users would expect certain things to behave. At the moment, the global value is practically unusable.

I have been looking at the e2e test failing. I honestly have no clue why that would fail.

```yaml
  - command: helm install trivy-operator ./deploy/helm --namespace trivy-system
      --create-namespace --set="operator.builtInTrivyServer=true"
      --set="image.tag=e2e" --set="targetNamespaces=e2e-test"
      --set="operator.sbomGenerationEnabled=true"
```

Looking at above code it relies on `image.tag` which hasn't been touched in this PR. Looking at the workflow logs are very minimal, so I have no clue why that is failing.

Looking at the `tests/e2e/client-server/workload` I don't see any obvious reason why it fails (new to `kuttl`). The workflow seems to indicate that the creation of that nginx pod is failing, but it is pretty vague in the logging on the workflow, so not sure.

Is the failure really only on this PR or is it something that happens on all PRs at the moment?

afdesk:
> Is the failure really only on this PR or is it something that happens on all PRs at the moment?

i've created a simple PR and the tests are passed without errors: https://github.com/aquasecurity/trivy-operator/pull/2654

but it doesn't matter for a while.

actually you araised an interesting question about node-collector.

I'm curious why this issue hasn't been raised earlier — perhaps there's a common workaround that most users are relying on? If that's the case, we should carefully consider whether introducing a breaking change is justified, especially if the existing workaround is sufficient for the community.

@simar7 wdyt?

marcofranssen:
> > Is the failure really only on this PR or is it something that happens on all PRs at the moment?
> 
> i've created a simple PR and the tests are passed without errors: #2654
> 
> but it doesn't matter for a while.
> 
> actually you araised an interesting question about node-collector.
> 
> I'm curious why this issue hasn't been raised earlier — perhaps there's a common workaround that most users are relying on? If that's the case, we should carefully consider whether introducing a breaking change is justified, especially if the existing workaround is sufficient for the community.
> 
> @simar7 wdyt?

Yes, as long you don't use the `global.image` all is fine.

Expected behavior:

- define global
- override per specific image where required

Actual behavior

- define global
- overrides per image don't work.

This PR fixes the Actual Behavior to behave as expected. + It normalizes the interface for all images in the chart.

simar7:
@marcofranssen - I'm a little bit concerned why the tests are failing. Also since this will be a breaking change, I'd like to understand more clearly why no user has encountered this yet as @afdesk mentioned. Using `global.image` seems like a pretty standard thing to do.

marcofranssen:
@afdesk @simar7 I have just rebased the PR on top of latest main. Let's see if the tests will succeed after the rebase (awaiting approval for workflows), if not I could use some help on figuring out why the tests fail.

I have some more ideas to harmanize the UX of Helm values after this PR gets in.

Would be great if one of the maintainers can help out to move this forward. I'm not super familiar with the testing framework used.

marcofranssen:
Did another rebase @afdesk @simar7 . Anyone of you guys able to help me out moving this forward? Need some help on the e2e tests.

afdesk:
@marcofranssen I took a quick look at the error
the e2e test doesn't get a report in client - server mode.
I also see, that it uses `global.imageRegistry`.
so I'd recheck this template https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/templates/trivy-server/

marcofranssen:
@afdesk I finally found it with help of claude.

I missed one place in the configmap to use the new values. See latest commit. Hope the tests now succeed.

afdesk:
@marcofranssen thanks for your efforts! we appreciate it!

now Simar is on vacation, so let's wait for his opinion pls

marcofranssen:
@simar7 @afdesk any update? 

For what it is worth. This is how all charts are doing it, so this would align the behavior with any other opensource chart people might be using (e.g.):

- spire
- prometheus
- grafana
- tempo
- loki
- cert-manager

simar7:
> Also since this will be a breaking change, I'd like to understand more clearly why no user has encountered this yet as @afdesk mentioned. Using global.image seems like a pretty standard thing to do.

Could you explain this a bit? Not to undermine the issue here but - correct me if I'm wrong @afdesk but I'm yet to see another report about this.


Also I'd like to understand how this would impact existing deployments. While in-principle I agree that we should override if defined or else default back to the global value set, just want to make sure we don't break backwards compatibility.

PR #2665, opened 2025-07-18
feat: allow for annotations to be passed from resources to reports

lukasboettcher:
## Description

Currently only resource labels from scanned resources can be included in the reports. Labels are identifying and have a character limit of 63 chars. For non-identifying metadata, annotations are a better fit without any of the constraints on labels.

### Our use-case

We send trivy reports to a webhook and want to include routiing information in the reports, which does not fit in labels. This information is already present on the resources as annotations. So we would like to include this information in the reports.

## Checklist


lukasboettcher:
@afdesk @simar7 would mind running the ci again and taking a look? linting errors should be fixed now.

simar7:
hi @lukasboettcher - thanks for the PR but I'm a little concerned about implementing this in the operator given potential performance degradation. Trivy operator has a many critical hot paths that could get impacted by adding more metadata such as annotations to the resources.

lukasboettcher:
thanks for taking a look! understandable concern. is there anything that you are especially worried about? as far as I can tell, if a user does not specify any additional annotations to be included, most operations introduced by this should be constant. the only critical area is the lookup of the resources annotations via `b.controller.GetAnnotations`:
```
annotationsSet := make(map[string]string)
// append matching resource annotations by config to report
kube.AppendResourceAnnotations(b.resourceAnnotationsToInclude, b.controller.GetAnnotations(), annotationsSet)
// append custom annotations by config to report
kube.AppendCustomAnnotations(b.additionalReportAnnotations, annotationsSet)
```

this could be wrapped in a condition, to only check if `len(resourceAnnotationsToInclude) > 0`. same thing could be done for the labels.

in general I think, being in control of the generated report's metadata is a useful feature.

lukasboettcher:
bump @afdesk @simar7

afdesk:
@lukasboettcher sorry for the delay.
Simar is on vacation right now, he'll take a look when he's back

PR #2670, opened 2025-07-25
fix: Do not error out on invalid rego policies if any

simar7:
## Description

If there are no results returned upon processing an invalid policy, it should not be treated as an error as there are simply no errors that have actually taken place.

## Checklist


afdesk:
LGTM!
I tested it with next options:
```sh
$ helm install trivy-operator aquasecurity/trivy-operator --create-namespace --namespace trivy-system \
     --set image.repository=afdesk/trivy-operator --set image.tag=fix-invalid-logic \
     --set operator.scanJobTTL=30m --set trivy.resources.limits.memory=1Gi
```

PR #2683, opened 2025-08-03
fix: send webhook reports when altstorage is used

### LINKED ISSUE #2682: Webhook reporting is not working when alternative storage is used

1. Install trivy-operator onto the cluster with the following values set:
```
operator:
  webhookBroadcastURL: "..." # any webhook here
alternateReportStorage:
  enabled: true
```

All types of reports (`VulnerabilityReport`,`ExposedSecretReport`,`ConfigAuditReport`,`InfraAssessmentReport`,`ClusterComplianceReport`,`RbacAssessmentReport`,`ClusterRbacAssessmentReport`,`ClusterConfigAuditReport`,`ClusterInfraAssessmentReport`,`SbomReport`) being sent to the webhook

Instead, the webhook is only getting `ClusterComplianceReport`

The issue occurs because:

1. When AltReportStorageEnabled is false (default):
* Reports are written as Kubernetes Custom Resources (CRs) in etcd
* The webhook reconciler watches these CRDs and sends webhook notifications when they change
* `WebhookBroadcastURL` receives notifications

2. When AltReportStorageEnabled is true:
* Reports are written to the filesystem as JSON files
* No CRs are created/updated
* The webhook reconciler has nothing to watch since no CRs are created
* `WebhookBroadcastURL` receives no notifications

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): `{"Version":"0.27.3","Commit":"a086c84b3f20ed1fe670c3c29621c4de32e57468","Date":"2025-07-10T22:29:22Z"}`
- Kubernetes version (use `kubectl version`):
```
Client Version: v1.33.3
Kustomize Version: v5.6.0
Server Version: v1.32.1
```
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): host k8s node is `Ubuntu 20`

vmax:
## Related issues
- Close #2682

## Checklist

### DIFF
--- a/pkg/compliance/clustercompliancereport.go
+++ b/pkg/compliance/clustercompliancereport.go
@@ -88,11 +88,7 @@ func (r *ClusterComplianceReportReconciler) generateComplianceReport(ctx context
 				log.Info("Cluster compliance report written", "path", reportPath)
 
 				// Send webhook notification for compliance report
-				if r.Config.WebhookBroadcastURL != "" {
-					if err := webhook.SendWebhookReport(&report, r.Config); err != nil {
-						log.Error(err, "Failed to send compliance report via webhook")
-					}
-				}
+				webhook.SendWebhookReport(&report, r.Config, log)
 
 				return nil
 			}
--- a/pkg/compliance/clustercompliancereport.go
+++ b/pkg/compliance/clustercompliancereport.go
@@ -88,7 +88,7 @@ func (r *ClusterComplianceReportReconciler) generateComplianceReport(ctx context
 				log.Info("Cluster compliance report written", "path", reportPath)
 
 				// Send webhook notification for compliance report
-				webhook.SendWebhookReport(&report, r.Config, log)
+				webhook.SendWebhookReport(reportData, r.Config, log)
 
 				return nil
 			}
--- a/pkg/compliance/clustercompliancereport.go
+++ b/pkg/compliance/clustercompliancereport.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aquasecurity/trivy-operator/pkg/ext"
 	"github.com/aquasecurity/trivy-operator/pkg/operator/etc"
 	"github.com/aquasecurity/trivy-operator/pkg/utils"
+	"github.com/aquasecurity/trivy-operator/pkg/webhook"
 )
 
 type ClusterComplianceReportReconciler struct {
@@ -85,6 +86,15 @@ func (r *ClusterComplianceReportReconciler) generateComplianceReport(ctx context
 					return err
 				}
 				log.Info("Cluster compliance report written", "path", reportPath)
+
+				// Send webhook notification for compliance report
+				if r.Config.WebhookBroadcastURL != "" {
+					if err := webhook.SendWebhookReport(&report, r.Config); err != nil {
+						log.Error(err, "Failed to send compliance report via webhook")
+					}
+				}
+
+				return nil
 			}
 			err = r.Mgr.GenerateComplianceReport(ctx, report.Spec)
 			if err != nil {
--- a/pkg/compliance/clustercompliancereport.go
+++ b/pkg/compliance/clustercompliancereport.go
@@ -88,9 +88,7 @@ func (r *ClusterComplianceReportReconciler) generateComplianceReport(ctx context
 				log.Info("Cluster compliance report written", "path", reportPath)
 
 				// Send webhook notification for compliance report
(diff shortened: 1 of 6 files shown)


pkg/compliance/clustercompliancereport.go
```diff
@@ -86,6 +87,13 @@ func (r *ClusterComplianceReportReconciler) generateComplianceReport(ctx context
 				}
 				log.Info("Cluster compliance report written", "path", reportPath)
 
+				// Send webhook notification for compliance report
+				if r.Config.WebhookBroadcastURL != "" {
+					if err := webhook.SendWebhookReport(&report, r.Config); err != nil {
+						log.Error(err, "Failed to send compliance report via webhook")
+					}
+				}
+
```
simar7: Thanks for the PR! Can you add a small unit test that would exercise this? One idea to do so would be to create an http test server and then pass its URL to this function. That way we can assert the results within the test server.

vmax: @simar7 added a test in https://github.com/aquasecurity/trivy-operator/pull/2683/commits/3ba57d23dde20f542cc480b386d1b7cb0f8b098e; plz review

vmax:
@simar7 I believe the linter should now be happy and tests are there — could you please do another round of reviews? :)

pkg/webhook/webhookreporter.go
```diff
@@ -103,6 +118,31 @@ func (r *WebhookReconciler) reconcileReport(reportType client.Object) reconcile.
...
+
+	resp, err := hc.Do(req)
+	defer func() {
+		if resp != nil {
+			_ = resp.Body.Close()
+		}
+	}()
+	if err != nil {
+		return fmt.Errorf("failed to send reports to endpoint: %w", err)
+	}
```
afdesk:
```suggestion
	if err != nil {
		return fmt.Errorf("failed to send reports to endpoint: %w", err)
	}
	if resp != nil {
		_ = resp.Body.Close()
	}
```

afdesk: I'm unsure the `if resp != nil` is needed, IIRC if there is no error, `resp` can't be nil. need to check it

afdesk: Also I think we need to log `status` and `message` from a response.
as for me, we shouldn't skip this information 

@simar7 WDYT?

pkg/webhook/webhookreporter.go
```diff
@@ -68,9 +68,24 @@ func (r *WebhookReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 
 	}
+
 	return nil
 }
 
+// SendWebhookReport sends a report directly via webhook without going through CRD reconciliation
+// This is used when AltReportStorageEnabled is true and reports are written to filesystem
+func SendWebhookReport(reportData []byte, config etc.Config, log logr.Logger) {
```
afdesk: I'd improve a name for this function.
something like `SendWebhookReportIfAvailable`, in favor of other functions in `config`: https://github.com/aquasecurity/trivy-operator/blob/main/pkg/plugins/trivy/config.go#L375

@simar7 WDYT?

afdesk (commented):
@vmax thanks for your cool PR.
I left a few small comments for discussion

kovasaurus:
bump, any progress with this?

pkg/webhook/webhookreporter.go
vmax: Applied, thanks — the response body is now closed via defer once the request succeeds.

vmax: You're right — when there's no error `resp` can't be nil. Reordered to check the error first, then `defer resp.Body.Close()` without the nil guard.

vmax: Agreed. Non-2xx responses now return an error containing the status line and response body, so the caller logs it instead of silently dropping that information.

pkg/webhook/webhookreporter.go
vmax: Good call — renamed to `SendWebhookReportIfAvailable` (it's a no-op when no webhook URL is configured) and updated all call sites.

vmax:
Rebased onto latest `main` and addressed @afdesk's feedback:

- Renamed `SendWebhookReport` → `SendWebhookReportIfAvailable` (no-op when no webhook URL is set).
- Webhook now checks the response status and returns an error with the status line + body on non-2xx replies instead of discarding it.
- Response body is closed only after a successful request (reordered, since `resp` is non-nil when there's no error).

Ready for another look 🙏

PR #2666, opened 2025-07-19
feat(dev): add support for pprof

simar7:
## Description
dev: add support for pprof

## Related issues
- Investigates https://github.com/aquasecurity/trivy-operator/issues/2541

## Checklist
- [x] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

PR #2693, opened 2025-08-06
feat: add support any kinds for policy checks

afdesk:
## Checklist

### DIFF
--- a/tests/itest/trivy-operator/behavior/behavior.go
+++ b/tests/itest/trivy-operator/behavior/behavior.go
@@ -69,6 +69,73 @@ func VulnerabilityScannerBehavior(inputs *Inputs) func() {
 
 		})
 
+		Context("When PersistentVolumeClaim is created", func() {
+
+			var ctx context.Context
+			var pvc *corev1.PersistentVolumeClaim
+
+			BeforeEach(func() {
+				ctx = context.Background()
+				qty := resource.MustParse("1Gi")
+				pvc = &corev1.PersistentVolumeClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: inputs.PrimaryNamespace,
+						Name:      "pvc-" + rand.String(5),
+					},
+					Spec: corev1.PersistentVolumeClaimSpec{
+						AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+						Resources: corev1.VolumeResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceStorage: qty,
+							},
+						},
+					},
+				}
+				err := inputs.Create(ctx, pvc)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("Should not create VulnerabilityReport", func() {
+				Consistently(inputs.HasVulnerabilityReportOwnedBy(ctx, pvc), time.Minute, inputs.PollingInterval).Should(BeFalse())
+			})
+
+			AfterEach(func() {
+				err := inputs.Delete(ctx, pvc)
+				Expect(err).ToNot(HaveOccurred())
+			})
+		})
+
+		Context("When unsupported resource Secret is created", func() {
+
+			var ctx context.Context
+			var secret *corev1.Secret
+
+			BeforeEach(func() {
+				ctx = context.Background()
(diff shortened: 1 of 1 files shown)


pkg/vulnerabilityreport/controller/workload.go
```diff
@@ -94,6 +94,10 @@ func (r *WorkloadController) SetupWithManager(mgr ctrl.Manager) error {
 	// Determine which Kubernetes workloads the controller will reconcile and add them to resources
 	targetWorkloads := r.Config.GetTargetWorkloads()
```
simar7, line 95: I believe we also should update `GetTargetWorkloads()` as well https://github.com/aquasecurity/trivy-operator/blob/6cf2d176262995ee777fa5c53ff88695393882bd/pkg/operator/etc/config.go#L133

pkg/operator/etc/config.go
```diff
@@ -54,7 +54,7 @@ type Config struct {
 	WebhookBroadcastTimeout                      *time.Duration `env:"OPERATOR_WEBHOOK_BROADCAST_TIMEOUT" envDefault:"30s"`
 	WebhookBroadcastCustomHeaders                string         `env:"OPERATOR_WEBHOOK_BROADCAST_CUSTOM_HEADERS"`
 	WebhookSendDeletedReports                    bool           `env:"OPERATOR_SEND_DELETED_REPORTS" envDefault:"false"`
-	TargetWorkloads                              string         `env:"OPERATOR_TARGET_WORKLOADS" envDefault:"Pod,ReplicaSet,ReplicationController,StatefulSet,DaemonSet,CronJob,Job"`
+	TargetWorkloads                              string         `env:"OPERATOR_TARGET_WORKLOADS" envDefault:"Pod,ReplicaSet,ReplicationController,StatefulSet,DaemonSet,CronJob,Job,PersistentVolume"`
```
simar7, line 57: If this is configurable today, why can't the user scan them now? I must be missing something.

simar7:
Just thinking out loud how to test this change: As part of our integration tests, we can specify the new PV resource and have it get scanned and create reports based on that. WDYT?

simar7:
Another point about testing, I added perf support on a feature branch (to enable profiling) but I think I'll enhance it by adding support for it to be enabled with a configuration option. That we the performance team can test this branch to see if there's a big performance overhead or not.

I updated that PR here https://github.com/aquasecurity/trivy-operator/pull/2666

pkg/vulnerabilityreport/controller/workload.go
afdesk: updated https://github.com/aquasecurity/trivy-operator/pull/2693/commits/445eae3f398c1d4f594b8f7c1c6545700188988b

pkg/operator/etc/config.go
afdesk: it was impossible because specific Kind values were hard-coded in multiple places. 
When the actual type didn’t match one of those hard-coded cases, the code returned an error.
for example:
https://github.com/aquasecurity/trivy-operator/blob/main/pkg/kube/object.go#L204-L237
https://github.com/aquasecurity/trivy-operator/blob/main/pkg/kube/object.go#L336-L385
and other

afdesk:
> Just thinking out loud how to test this change: As part of our integration tests, we can specify the new PV resource and have it get scanned and create reports based on that. WDYT?

tried to add it here https://github.com/aquasecurity/trivy-operator/pull/2693/commits/37dd96d5a404f559cbc349d3ca1af8816f352a7c

tests/itest/trivy-operator/behavior/behavior.go
```diff
@@ -456,6 +457,79 @@ func ConfigurationCheckerBehavior(inputs *Inputs) func() {
 			})
 
 		})
+
```
simar7, line 529: Can we also add another case where we don't specify the scanning of a PVC but create one? In such a case the assertion should ensure that no report is generated.

Another case should be that we create a new resource (something other than what we support by default) it shouldn't generate any reports either.

afdesk: added https://github.com/aquasecurity/trivy-operator/pull/2693/commits/baf47eea6b71c256231761b00c34adcbc2f0d523

PR #2713, opened 2025-08-18
fix: Ensure configFile volume is mounted for initContainer

### LINKED ISSUE #2711: Operator self-scan job enters Error state after applying fix for configFile volume mount
### Overview

This issue tracks a bug that prevents the trivy.configFile feature from working correctly in Standalone mode. The root cause is that the initContainer in scan jobs does not receive the configFile volume mount, as the volume is defined too late in the pod spec creation process.

The fix, detailed in our Pull Request #2713 , resolves this by ensuring the volume is defined before the initContainer is created.

During our debugging, we encountered a secondary issue where the operator's self-scan would fail when using a locally built image. We have since determined this to be specific to local testing, and not a bug in the operator itself. We believe this "issue" could have been solved by reading the docs more thoroughly.

This work is also related to the configFile improvements in #2702.

### Environment
Trivy-Operator Version: Based on 0.28.0 (running a custom dev build from a fork - chart version 0.30.0).

Kubernetes Version: v1.33.1

Cluster Environment: KIND 

### Details
**Initial Goal:** Our primary objective was to use the trivy.configFile Helm value to apply a VEX document for vulnerability suppression.

**The configFile Bug**: We discovered that this was not possible because the path specified by trivy.configFile (e.g., /etc/trivy/trivy-config.yaml) was never created inside the scan job pod. This led us to find that in pkg/plugins/trivy/image.go, the initContainer is defined before the GenerateConfigFileVolumeIfAvailable() function is called, thus preventing the volume from being mounted.

**The Fix**: Our PR corrects this by moving the GenerateConfigFileVolumeIfAvailable() call to a position before the initContainers block is defined. With this fix, the configFile is correctly mounted, and we verified that subsequent scan jobs (e.g., on an nginx workload) successfully applied our VEX suppressions.

**Investigating a Secondary Symptom**: After applying the fix, we observed that the operator's initial self-scan would consistently fail, with the main container entering an Error state. To isolate this, we performed two tests:

**Test A:** We cloned the upstream main branch, built a new image with no code changes, and deployed it. The self-scan failed with the exact same error.

**Test B:** We configured the deployment to use the official public image from the registry. The self-scan completed successfully.

### Conclusion 

The code change in our PR #2713  fixes the bug preventing the configFile feature from working.

bananasplit393:
Fixes #2711 

This PR resolves a bug where the configFile feature is not properly implemented in the scan jobs created in Standalone mode.

Currently, the initContainer spec is defined before the GenerateConfigFileVolumeIfAvailable() function is called. As a result, the volume mount for the configFile is never added to the initContainer, preventing it from accessing any custom configuration.

This change moves the GenerateConfigFileVolumeIfAvailable() call to a position before the initContainers slice is defined. This ensures that the initContainer correctly receives the volume mount, making the configFile feature fully functional for all containers within the scan job pod.

Tests have been updated to mirror the code change.


trivy-operator
afdesk, line 1: do we really need this file?

afdesk:
@bananasplit393 thanks for your time and efforts
I left a comment to clarify the PR

trivy-operator
ReneDiff: No, we do not. I do not know how it got through... It is removed now, thank you.

ReneDiff:
> @bananasplit393 thanks for your time and efforts I left a comment to clarify the PR

Hey, we just realized that our fork was 4 commits behind main, so we merged again - so we need a new review... 😅

afdesk:
thank you guys!

PR #2726, opened 2025-08-27
perf: optimize memory configuration for scan jobs

simar7:
## Description

- Increase memory requests from 100M to 200M for better performance
- Increase memory limits from 500M to 1Gi to handle large scans

## Related discussions
- https://github.com/aquasecurity/trivy-operator/discussions/2674

## Checklist

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -498,11 +498,11 @@ trivy:
   resources:
     requests:
       cpu: 100m
-      memory: 100M
+      memory: 200M  # Increased for better performance
       # ephemeralStorage: "2Gi"
     limits:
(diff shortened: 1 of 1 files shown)


itaysk:
I think the comments are unnecessary. can you share what led you to increasing the limits? do we recommend this for all new installations?

afdesk:
> can you share what led you to increasing the limits?

@itaysk actually I always get `oomkilled` with the current settings on default KinD instances.
so I'd like to increase a memory limit too

afdesk (changes requested):
it seems we have to run `mage generate:manifests` to update `deploy/static/trivy-operator.yaml`

PR #2746, opened 2025-09-15
chore: bump up common k8s deps

afdesk:
## Description
In controller-runtime version v0.22+, several breaking changes were introduced that affected our test setup:
* Fake client no longer supports objects with pointer ObjectMeta, and it clears TypeMeta for structured objects:
- https://github.com/kubernetes-sigs/controller-runtime/pull/3228
- https://github.com/kubernetes-sigs/controller-runtime/pull/3229

As a result, some of our existing test data became incompatible.

This PR addresses these issues by:
* Removing TypeMeta from test fixtures.
* Updating report type resolution: since the fake client may no longer set GVK on Get, we now rely on the Go type instead of GVK to determine the report type.

Additionally, due to the breaking changes, the hash of the test resources has also been affected. To make the tests more resilient, the hash value is no longer hardcoded as a constant but is now computed dynamically.

It fixes fails in https://github.com/aquasecurity/trivy-operator/pull/2749

## Checklist


pkg/vulnerabilityreport/builder_test.go
```diff
@@ -105,6 +106,14 @@ func TestScanJobBuilder(t *testing.T) {
 			Get()
 		g.Expect(err).ToNot(gomega.HaveOccurred())
 		g.Expect(job).ToNot(gomega.BeNil())
+
+		specHash := kube.ComputeHash(corev1.PodSpec{
+			Containers: []corev1.Container{{
+				Name:  "nginx",
+				Image: "nginx:1.16",
+			}},
+		})
```
simar7: While this makes the tests resilient like you mentioned, it also allows the tests to silently pass, even if there's an actual problem because the hash will never be statically defined and always get computed on the fly.

I think it's better to have the hash statically defined, similar to a static golden file. If the tests break and we have to update them often, we should re-think what we are asserting as unit tests for the most part should offer resilience across code changes.

WDYT?

afdesk: it makes sense!
done https://github.com/aquasecurity/trivy-operator/pull/2746/commits/5bc396b191a5c80dce9d2a1e5c63b97f3a7772da

simar7 (approved):

PR #2750, opened 2025-09-17
feat: add support for custom Trivy ignore file

### LINKED ISSUE #2744: Support trivyignore.yaml file
We want to support to ignore CVE's in all container images for a certain binary, like `gosu`, and I currently can't get it working in trivy-operator without having an "own" trivy image where I have added the `trivyignore.yaml`.

The current `helm` installation (and I think the operator to) doesn't support the `trivyignore.yaml` format, only the more "basic" ignore all CVE's independent of what binary they are found in etc. 

The OPA policy doesn't have the context of the binary so it's not possible to only whitelist for one binary in an image.

*Current workaround:*
Build an own image like:
```
FROM mirror.gcr.io/aquasec/trivy:0.65.0
COPY trivy_ignore_rules.yaml /root/.trivyignore.yaml
ENV TRIVY_IGNOREFILE=/root/.trivyignore.yaml
```

sample `trivy_ignore_rules.yaml`
```
vulnerabilities:
  # gosu vulnerabilities can be ignored since it does use affected functionality
  # gosu is scanned every week (see: https://github.com/tianon/gosu/blob/master/.github/workflows/ci.yml) and if workflow does not fail it has no new vulnerabilities
  - id: CVE-XXX-YYYY
    paths:
      - usr/local/bin/gosu
    statement: Library not reached from the code
```

Point to the image in the helm installation in:
`trivy.image.repository`

afdesk:
## Description
This PR fixes Helm rendering of `trivy.ignoreFile` and adds support for a custom ignore file name via `trivy.ignoreFileName`.

## Reproduction steps
```sh
$ helm install trivy-operator ./deploy/helm --create-namespace --namespace trivy-system --values trivyignore.yaml
$ kubectl run kube-bench --image aquasec/kube-bench:v0.11.1
$ kubectl get vulns -o yaml | grep CVE-2025-47907
```

<details><summary>trivyignore values</summary>

```yaml
image:
  registry: "docker.io"
  repository: "afdesk/trivy-operator"
  tag: "ignorefilename"
operator:
  scanJobTTL: 30m

trivy:
  ignoreFileName: .trivyignore.yaml
  ignoreFile:
    vulnerabilities:
     - id: CVE-2025-47907
       statement: There is no fix yet
  timeout: "15m0s"
  resources:
    limits:
       memory: 1Gi
```
</details> 

Before:
```sh
Error: INSTALLATION FAILED: template: trivy-operator/templates/configmaps/trivy.yaml:86:18: executing "trivy-operator/templates/configmaps/trivy.yaml" at <4>: wrong type for value; expected string; got []interface {}
```
After - `trivy-operator` deployed successfully.

## Related issues
- Close #2744

## Checklist

### DIFF
--- a/.github/workflows/chart-testing.yaml
+++ b/.github/workflows/chart-testing.yaml
@@ -83,6 +83,9 @@ jobs:
       - name: Setup chart-testing
         id: lint
         uses: helm/chart-testing-action@v2.7.0
+        with:
+          # v6.0.0 resolved the compatibility issue with Python > 3.13. may be removed after the action itself is updated
+          yamale_version: "6.0.0"
 
       - name: Run chart-testing
         run: ct lint-and-install --validate-maintainers=false --charts deploy/helm
--- a/.github/workflows/publish-helm-chart.yaml
+++ b/.github/workflows/publish-helm-chart.yaml
@@ -41,6 +41,9 @@ jobs:
       - name: Setup Chart Linting
         id: lint
         uses: helm/chart-testing-action@v2.7.0
+        with:
+          # v6.0.0 resolved the compatibility issue with Python > 3.13. may be removed after the action itself is updated
+          yamale_version: "6.0.0"
 
       - name: Setup Kubernetes cluster (KIND)
         uses: helm/kind-action@v1.12.0 # v1.5.0


simar7 (commented):
@afdesk what's the motivation to add the option for a custom file name for a trivy ignore file? This behaviour will be different from what we have commercially or even in the Trivy OSS CLI and to me feels unnecessary unless there's a good reason that I might have missed.

afdesk:
> what's the motivation to add the option for a custom file name for a trivy ignore file?

`.trivyignore` and `.trivyignore.yaml` have a different behavior in Trivy.
the first is the old style - only string
the second accepts yaml
now `.trivyignore` is a hard-coded constant, so there is no way to select a type of ignore file.

I've added a new option because i dont want to create a breaking change.

tests/itest/trivy-operator/trivyignore_test.go
```diff
@@ -0,0 +1,101 @@
...
+		if !strings.Contains(current, IgnoredCVE) {
+			if current != "" && !strings.HasSuffix(current, "\n") {
+				current += "\n"
+			}
+			current += fmt.Sprintf("%s\n", IgnoredCVE)
+		}
+		cm.Data["trivy.ignoreFile"] = current
+
+		By("Updating trivy.ignoreFile in ConfigMap")
+		Expect(kubeClient.Update(ctx, cm)).To(Succeed())
```
simar7: I'm just curious why this logic is needed? Shouldn't a test have no prior artifacts in the setup?

afdesk: oh, sure - removed https://github.com/aquasecurity/trivy-operator/pull/2750/commits/e2ef4212c5bab39f8dd6e38a57f0793b44787c24

PR #2788, opened 2025-10-17
fix: delete completed scan jobs in any case

### LINKED ISSUE #2785: Scan jobs never deleted on failures to write to etcd

1. Deployed a container with a massive number of vulnerabilities, causing a vulnerability report that is too large to write to etcd.
2. Operator fails to write to etcd (as expected), but scan job never goes away.

Scan job is cleaned up despite failure to write to etcd.

scanJobTTL could have been used as a workaround, but now all scan jobs are respecting the TTL set, so it would make the normal jobs hang out for that length of time as well, taking up scanning slots much longer than they need to.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.29.0
- Kubernetes version (use `kubectl version`): 1.31.4
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): RHEL9

afdesk:
## Description

This PR slightly adjusts the operator’s behavior when errors occur during the processing of a completed scan job.

Trivy-operator will now delete the scan job if any errors are encountered during its processing.

This change is intended to prevent scan jobs from remaining undeleted indefinitely in error scenarios.

**Note:** When `scanJobTTL` is configured, the scan job deletion will respect this TTL setting.

## Related issues
- Close #2785

## Checklist

### DIFF
--- a/pkg/vulnerabilityreport/builder_test.go
+++ b/pkg/vulnerabilityreport/builder_test.go
@@ -580,6 +580,104 @@ func TestScanJobBuilder(t *testing.T) {
 			},
 		}))
 	})
+
+	// TTL behavior: should respect TTL when set (>0), and omit when 0 or nil
+	t.Run("Should set TTLSecondsAfterFinished when TTL > 0", func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		ttl := 2 * time.Minute
+		job, _, err := vulnerabilityreport.NewScanJobBuilder().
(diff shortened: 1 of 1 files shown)


afdesk:
> Can we add a test for this? A special case to handle in that test would be if `ttl` is set, it should respect that.

done https://github.com/aquasecurity/trivy-operator/pull/2788/commits/b60f12f53ca9275e6e7536b3d4f24f497ee5e53f

PR #2811, opened 2025-11-15
fix: When submitting a scan job correctly check presence of reused reports before fetching credentials for each image

### LINKED ISSUE #2809: No private registry credentials for containers being analyzed by jobs where at least one image has a ClusterSbomReport

Running `trivy-operator` in a kubernetes cluster with `sbom` cache enabled. After some scans some `ClusterSbomReports`  are present. Consequently, when `trivy-operator` scans a pod with more that one container where one image that does not have a corresponding `ClusterSbomReport` and is from a private registry, and at least one image has a corresponding `ClusterSbomReport`, the generated pod spec does not include private registry credentials, causing the scan to fail for that image.

Private registry credentials being correctly generated for the container running the private registry image.

After reading the code for a bit I believe the problem is related to [this line of code](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/vulnerabilityreport/controller/workload.go#L258). 

**Environment:**

- Trivy-Operator version (use `trivy-operator version`):  0.28.0
- Kubernetes version (use `kubectl version`): v1.32.7
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Debian 12

festeveira:
## Description
This PR fixes an issue in the logic that determines when to fetch private registry credentials. Previously, the code only checked whether the number of reused reports was greater than 0. This caused issues when the workload was composed of several images and at least one of the images had a corresponding `ClusterSbomReport` which was reused for the current scan; if one of the remaining images did not have a corresponding `ClusterSbomReport` and was from a private registry, it's credentials were not being fetched and added to the `PodSpec`'s environment.

The updated logic instead compares the number of reused reports with the total number of container images. If the counts differ, it indicates that not all images have reusable reports, so the system will fetch the necessary private registry credentials for each image. 

I have tested this change in my environment and it seemed to fix the related issue.

## Related issues
- Close #2809

## Checklist


afdesk:
@festeveira thanks for your efforts!
can we combine this PR with #2813? it looks like a similar issues.
also Could you pls add test cases for it? i realize there are no tests here, but it'd be better to have it

PR #2839, opened 2025-12-10
feat: add support for cache-backend setting

### LINKED ISSUE #2769: Multiple container scans fail with database lock error when trivy-operator is configured to use trivy v0.66.0

When a vulnerability scan job contains multiple containers, only the first container is successfully scanned. Subsequent containers fail with a database lock error shown below, preventing completion of the full scan job.

> Fatal error  run error: init error: DB error: error in vulnerability DB initialize: vulnerability database may be in use by another process: timeout

I believe this is caused by https://github.com/aquasecurity/trivy/pull/9307

All containers in a multi-container scan job should be scanned successfully without database conflicts.

Downgrading to `trivy.image.tag: 0.65.0` resolves the issue and allows all containers to be scanned successfully.

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.29.0
- Kubernetes version (use `kubectl version`): 1.33
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Linux

henkka:
Fixes #2769


afdesk:
Thanks for your efforts!
I'm unsure about this way to pass Trivy options.

@simar7 WDYT?

Issue #757, opened 2022-12-05
Reconciler error: Large size VulnerabilityReports will failed create and remains scan jobs


mrtc0:
**What steps did you take and what happened:**

Creation of VulnerabilityReports fails if the number of detected vulnerabilities is too large.

**Steps to reproduce**

1. Create a Pod with `python:3.5.10-buster` image on a cluster with trivy-operator installed

(`python:3.5.10-buster` has 1252 vulnerabilities.)

```shell
$ kubectl create deployment python -n default --image python:3.5.10-buster
```

3. Scan finishes successfully, but creation of VulnerabilityReports fails

```
# Scan is completed successfully, but no VulnerabilityAlerts are created and no jobs are deleted
❯ kubectl get jobs -n trivy-system
NAME                                  COMPLETIONS   DURATION   AGE
scan-vulnerabilityreport-7d8db4fc84   1/1           73s        5m42s

❯ kubectl get pods -n trivy-system
NAME                                        READY   STATUS      RESTARTS   AGE
scan-vulnerabilityreport-7d8db4fc84-gsztw   0/1     Completed   0          4m12s
trivy-operator-54bc4769db-5djnd             1/1     Running     0          85m

❯ kubectl -n trivy-system logs scan-vulnerabilityreport-7d8db4fc84-gsztw | tr -d "\n" | base64 -d | bunzip2 | jq .ArtifactName
Defaulted container "python" out of: python, 95164937-7089-491c-ab58-7f52bc8a8cce (init)
"python:3.5.10-buster"

❯ kubectl get vulnerabilityreports -n default
No resources found in default namespace.

# Reconciler error is logged in Operator
❯ kubectl logs -n trivy-system trivy-operator-54bc4769db-5djnd
...
[...]

PR #2854, opened 2026-01-14
feat: allow to remove fields from vulnerability reports and omit empty fields

### LINKED ISSUE #442: Reduce the amount of vulnerability fields stored in the vulnerabilities report
in some clusters the vulnerability report produce by 'trivy-operator' is hitting the `max request limit of etcd`

it is suggested to reduce the amount of fields stored for each vulnerability , that same data already exist in [AVD](https://avd.aquasec.com/) so the data will not be lost.

example of fields that can be removed: `title` , `fixed version` and etc

Vulnerability data with fields :
```
{"vulnerabilityID":"CVE-2022-1304","resource":"libcom","installedVersion":"1.43.4-2+deb9u1","fixedVersion":"","severity":"HIGH","title":"e2fsprogs: out-of-bounds read/write via crafted filesystem","primaryLink":"https://avd.aquasec.com/nvd/cve-2022-1304","links":[],"score":5.8,"target":""}
```
Vulnerability data without fields :
```
{"vulnerabilityID":"CVE-2022-1304","resource":"libcom","installedVersion":"1.43.4-2+deb9u1","severity":"HIGH","primaryLink":"https://avd.aquasec.com/nvd/cve-2022-1304\"}
```

The removed fields can be found in AVD for this [CVE-2022-1304](https://avd.aquasec.com/nvd/cve-2022-1304) so no data get lost

sathieu:
## Description

I propose two changes to limit the problems of etcd max size reached (#757):

- allow to remove some fields from the vulnerability report
- omit (now) empty fields from produced resource

Before:

```yaml
  vulnerabilities:
  - fixedVersion: 0.45.0
    installedVersion: v0.43.0
    lastModifiedDate: "2025-12-11T19:36:41Z"
    links: []
    packagePURL: pkg:golang/golang.org/x/crypto@v0.43.0
    primaryLink: https://avd.aquasec.com/nvd/cve-2025-47914
    publishedDate: "2025-11-19T21:15:50Z"
    resource: golang.org/x/crypto
    score: 5.3
    severity: MEDIUM
    target: ""
    title: 'golang.org/x/crypto/ssh/agent: in golang.org/x/crypto/ssh/agent'
    vulnerabilityID: CVE-2025-47914
```

After first commit (and `trivy.additionalVulnerabilityReportFields=-Resource,-InstalledVersion,-FixedVersion,-PublishedDate,-LastModifiedDate,-Title,-PrimaryLink,-Score,-PURL`):

```yaml
  - fixedVersion: ""
    installedVersion: ""
    lastModifiedDate: ""
    publishedDate: ""
    resource: ""
    severity: MEDIUM
    target: ""
    title: ""
    vulnerabilityID: CVE-2025-47914
```

After both commits (and `trivy.additionalVulnerabilityReportFields=-<same>`):

```yaml
  - severity: MEDIUM
    vulnerabilityID: CVE-2025-47914
```

Notes:

- CRDs are changed, this needs to be added in release notes, as the CRD should be updated first
- vulnerabilityID can't be removed. I can change this if you want
- I haven't implemented https://github.com/aquasecurity/trivy-operator/issues/442#issuecomment-2338782160, as this breaks vulnid metrics
- I've updated tests

## Related issues
- Fixes #442
- Related  to #757, but not closing as there is still a limit

## Checklist


sathieu:
@simar7, @afdesk and others Please review 🙏 .

afdesk:
Hi @sathieu 
Thanks for your efforts! 
I'll take a look at this PR tonight or on Monday

pkg/plugins/trivy/config.go
```diff
@@ -80,28 +80,50 @@ type Config struct {
...
-	addFields := vulnerabilityreport.AdditionalFields{}
+	addFields := vulnerabilityreport.NewDefaultAdditionalFields()
 
 	fields, ok := c.Data[keyTrivyAdditionalVulnerabilityReportFields]
 	if !ok {
 		return addFields
 	}
 	for _, field := range strings.Split(fields, ",") {
+		value := !strings.HasPrefix(field, "-")
+		field = strings.TrimPrefix(field, "-")
```
afdesk: i think we should run `strings.TrimSpace` before these calls to support a case `  -PublishedDate`.

afdesk: also it'd be great to have a test for this case.

pkg/plugins/trivy/config_test.go
```diff
@@ -97,7 +97,26 @@ func TestConfig_GetAdditionalVulnerabilityReportFields(t *testing.T) {
...
 			}},
-			additionalFields: vulnerabilityreport.AdditionalFields{Description: true, Links: true, CVSS: true, Class: true, PackageType: true, PkgPath: true, Target: true},
+			additionalFields: vulnerabilityreport.AdditionalFields{
+				// default
+				Resource:         true,
+				InstalledVersion: true,
+				FixedVersion:     true,
+				PublishedDate:    true,
+				LastModifiedDate: true,
+				Severity:         true,
```
afdesk: there is a concern about `Severity`. it may affect on `Summary`.
so I suggest don't allow to exclude it. because I don't know a case when we need to exclude `Severity`. may be I miss something 
WDYT?

pkg/vulnerabilityreport/io.go
```diff
@@ -129,41 +129,76 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.ObjectRef) ([]v
...
+	Resource         bool
+	InstalledVersion bool
+	FixedVersion     bool
+	PublishedDate    bool
+	LastModifiedDate bool
+	Severity         bool
+	Title            bool
+	PrimaryLink      bool
+	Score            bool
+	PURL             bool
```
afdesk: about PURL. 
It seems `PURL` is too common name, because actually we can manage `PkgPURL` only, can't we?

afdesk (changes requested):

afdesk:
@sathieu Thank you very much for your effort and for submitting this PR — it’s greatly appreciated.

I’ve left a few inline comments to discuss some ideas.

I also have a small concern regarding backward compatibility. In particular, clients that previously relied on the presence of certain fields (even when they were empty) now need to be prepared for those fields to be omitted entirely.

Specifically, the handling of nil vs empty slices has changed: `Links` is no longer set when `sr.References == nil`. Previously, some code paths appeared to expect an empty slice ([]) instead.

This is likely not a critical issue, but it would be good to get feedback from the community and confirm that this change won’t cause unexpected behavior for existing consumers.

pkg/plugins/trivy/config.go
sathieu: Added strings.TrimSpace before removing prefix, with various tests: `" -Resource,- InstalledVersion,   -  FixedVersion, - PublishedDate,[...]"`

pkg/plugins/trivy/config_test.go
sathieu: `Severity` is now always kept (also in CRD, no omitempty).

pkg/vulnerabilityreport/io.go
sathieu: Changed to `PkgPURL`. Thanks

sathieu (commented):
Thanks @afdesk for your review.

About backward compatibility, we can use a new apiVersion (v1alpha2 or v1beta1 or v1), this will give safer approach to backward compatibility issues. This is documented [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/). A webhook will be needed to convert resources on-fly.

sathieu:
@afdesk I have extracted the non-breaking bits in #2860.

Please tell what I should do for the omitempty. I think adding a new apiVersion would be the best path. What do you think about v1beta1?

simar7:
> I also have a small concern regarding backward compatibility. In particular, clients that previously relied on the presence of certain fields (even when they were empty) now need to be prepared for those fields to be omitted entirely.

> 
> About backward compatibility, we can use a new apiVersion (v1alpha2 or v1beta1 or v1), this will give safer approach to backward compatibility issues. This is documented [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/). A webhook will be needed to convert resources on-fly.

@sathieu I think we certainly need to version this. It will be backwards breaking change to my understanding (removing existing fields). As you mentioned, we also will need a webhook to convert the resources on the fly.

sathieu:
@afdesk @simar7 This PR starts to be too big. Here is what I have done:

- feat: allow to remove fields from vulnerability reports (#2860)
- feat: reduce VulnerabilityReport size by omitting empty fields (i.e. v1beta1)
- feat: conversion between v1alpha1 et v1beta1
- feat: conversion webhook
- feat: webhook cert configuration
- :construction_worker: TODO inject certificate

I don't know how to handle the last item. Proposal:

- make the CRDs templatable in the Helm chart (by moving from crds to templates)
- mount secret in the deployment
- support various ways to inject certificate:
    - cert-manager (recommended): create a selfsigned Issuer, a certificate and inject in CRD (`cert-manager.io/inject-ca-from`)
    - generate using `genSelfSignedCert` (default)
    - pre-existing secret
    - disable webhook (i.e no conversion, and no mount in the deployment)

WDYT? Can we mandate cert-manager for the static installation? How to handle OLM installs?

PR #2856, opened 2026-01-15
fix: support metrics collection with alternate report storage

### LINKED ISSUE #2610: Alternate Report Storage doesn't provides metrics

I did install the latest operator with PVC storage option `alternateReportStorage.enabled: true`, then, I've deleted all reports stored in etcd.
Perhaps I'm wrong, but since the upgrade (a very welcome feature ! ) I don't have any metrics from the operator

The storage change would have been without side effects

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.27
- Kubernetes version (use `kubectl version`): 1.33.1
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Fedora CoreOS

rzala:
## Description

Fixes #2610

This PR adds support for Prometheus metrics collection when alternate report storage (filesystem-based) is enabled. Previously, enabling `alternateReportStorage.enabled: true` would cause all metrics like `trivy_image_vulnerabilities` to stop working because the metrics collector only read from Kubernetes CRDs.

## Changes

- **Added `StorageReader` interface** to abstract storage backend operations
- **Implemented `CRDStorageReader`** for reading from Kubernetes CRDs (existing default behavior)
- **Implemented `FilesystemStorageReader`** for reading from alternate storage filesystem
- **Updated `ResourcesMetricsCollector`** to use the `StorageReader` abstraction
- **Added comprehensive unit tests** for both storage backends and edge cases

## Impact

- ✅ **Backward compatible** - No breaking changes, CRD-based metrics work exactly as before
- ✅ **Dual-mode support** - Metrics now work with both CRD and filesystem storage
- ✅ **All report types supported** - VulnerabilityReport, ExposedSecretReport, ConfigAuditReport, RbacAssessmentReport, InfraAssessmentReport, ClusterComplianceReport
- ✅ **Production ready** - Robust error handling, logging, and graceful degradation

## Testing

Added comprehensive unit tests in `pkg/metrics/storage_reader_test.go`:
- Tests for CRD-based collection (validates backward compatibility)
- Tests for filesystem-based collection
- Edge case testing (missing directories, corrupt files, permissions)
- Backend selection tests

### With CRD Storage (default):
1. Deploy trivy-operator with default configuration
2. Verify metrics endpoint returns data

### With Alternate Storage:
1. Enable alternate storage: `alternateReportStorage.enabled: true`
2. Configure PVC-based storage
3. Verify metrics endpoint still returns data
4. Confirm metrics match report contents

## Checklist

- [x] Follows conventional commit format
- [x] References issue #2610
- [x] Includes comprehensive unit tests
- [x] Maintains backward compatibility
- [x] No changes to CRD definitions
- [x] No changes to metrics output format

### DIFF
--- a/pkg/configauditreport/controller/resource.go
+++ b/pkg/configauditreport/controller/resource.go
@@ -336,7 +336,7 @@ func (r *ResourceController) writeAlternateReports(resource client.Object, misCo
 				reportBuilder.ReportTTL(r.Config.ScannerReportTTL)
 			}
 
-			var configReport interface{}
+			var configReport any
 			var err error
 			if kube.IsClusterScopedKind(workloadKind) {
 				configReport, err = reportBuilder.GetClusterReport()
@@ -372,7 +372,7 @@ func (r *ResourceController) writeAlternateReports(resource client.Object, misCo
 				infraReportBuilder.ReportTTL(r.Config.ScannerReportTTL)
 			}
 
-			var infraReport interface{}
+			var infraReport any
 			var err error
 			if kube.IsClusterScopedKind(workloadKind) {
 				infraReport, err = infraReportBuilder.GetClusterReport()
@@ -408,7 +408,7 @@ func (r *ResourceController) writeAlternateReports(resource client.Object, misCo
 				rbacReportBuilder.ReportTTL(r.Config.ScannerReportTTL)
 			}
 
-			var rbacReport interface{}
+			var rbacReport any
 			var err error
 			if kube.IsClusterScopedKind(workloadKind) {
 				rbacReport, err = rbacReportBuilder.GetClusterReport()
--- a/pkg/configauditreport/controller/resource.go
+++ b/pkg/configauditreport/controller/resource.go
@@ -241,7 +241,7 @@ func (r *ResourceController) reconcileResource(resourceKind kube.Kind) reconcile
 		if r.Config.AltReportStorageEnabled && r.Config.AltReportDir != "" {
 			// Write reports to alternate storage if enabled
 			log.V(1).Info("Writing config, infra and rbac reports to alternate storage", "dir", r.Config.AltReportDir)
-			return r.writeAlternateReports(resource, misConfigData, log)
+			return r.writeAlternateReports(resource, misConfigData, resourceHash, policiesHash, resourceLabelsToInclude, additionalCustomLabel, log)
 		}
 		// create config-audit report
 		if !kube.IsRoleTypes(kube.Kind(kind)) || r.MergeRbacFindingWithConfigAudit {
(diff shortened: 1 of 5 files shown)


Sydney-MF:
Any idea on when this PR would get merged? Thanks:)

afdesk:
@Sydney-MF thanks for the ping
I'll take a look tomorrow

Sydney-MF:
@afdesk Any updates on this?:) really appreciate you for taking the time to take a look at this by the way

afdesk:
> @afdesk Any updates on this?:) really appreciate you for taking the time to take a look at this by the way

i'm on it right now
could you pls fix linter errors?

Sydney-MF:
> i'm on it right now could you pls fix linter errors?

It's not my PR so I'm not so sure if I should just go about it and modify it 😅

rzal:
Will have a look at the linting errors soon

afdesk:
Hi guys again! thanks for your efforts and contribution! I need more time to investigate it.

I have some concerns about performance here - StorageReader re-reads folders very ofter.
We have to evaluate it

Sydney-MF:
> Hi guys again! thanks for your efforts and contribution! I need more time to investigate it.
> 
> I have some concerns about performance here - StorageReader re-reads folders very ofter. We have to evaluate it

No worries, please keep us updated since there's quite an issue with the [etcd limit issue](https://github.com/aquasecurity/trivy-operator/discussions/2570) so the only way to make Trivy Operator work was for us to use alternateStorage, but then we lose the ability to see the metrics which is a big part of this tool's feature. But once again thank you for taking the time to take a look at this PR:)

Also thank you @rzala @rzal for the PR.

rzala:
@Sydney-MF thanks for the interest and for nudging this along!

@afdesk linting errors have been fixed in the latest push. Regarding the performance concern around frequent directory re-reads — happy to look into adding a file-watcher or in-memory cache with TTL if that's the preferred direction. Let me know what approach you'd like and I'll update the PR.

baznikin:
Maybe fine-tune this feature if disk scanning really became a problem? We're evaluating Trivy operator and only way we can run it is 1h ttl on CRD resources and limiting targets and scanners to very narrow set.

lukaspj:
We are seeing a lot of issues in our installation where the load on our API Server caused by Trivy is tremendous. (to the point of crashing our api server and causing incidents).

We are very dependent on Trivy's metrics so this PR would be amazing and solve a lot of headaches for us.

However, we have a lot of very big clusters and I'm also concerned about the performance of this, in one of our larger clusters we would have around:

 * 4000 SBOMReports
 * 3800 ConfigAuditReports
 * 700 VulnerabilityReports
 * Etc

In total it would be around 10195 resources, that could take quite some time to scan through using the filesystem like now. I don't know exactly what the solution is, I guess the preferable solution would be rewriting the alternativeReportStorage solution to use something like SQLite to make more efficient use of the disk.

matthiasbaldi:
@lukaspj we saw over the last few weeks similar issues on two of our clusters but with even less resources.
So that could also happen on less "powered" clusters or on clusters with a shared etcd in the background (sometimes on KaaS products).
And because it is like blocking the etcd that could have massive impact if you use flux or ArgoCD and the running rollouts.

May as workaround for CD tools: You can also reduce the load when you exclude resources like `aquasecurity.github.io/ClusterComplianceReport` in flux or ArgoCD setting.

But I look forward to get this PR here approved/merged and that we can test it like @baznikin wrote.

danielrlego:
@afdesk hi, any updates on this PR? thank you very much

stingalleman:
@afdesk Hi, any progress? :)

mprenditore:
it would be great if this can be merged, it can solve lot of problems on our cluters as well!

PR #2796, opened 2025-10-31
feat: Added possibility to configure TMPDIR environment variable on trivy s…

festeveira:
…can jobs

## Related issues
- Close #2795

Remove this section if you don't have related PRs.

## Checklist
- [X] I've read the [guidelines for contributing](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [X] I've updated the [documentation](https://github.com/aquasecurity/trivy-operator/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).

Issue #2795, opened 2025-10-31
It is currently impossible to configure TMPDIR environment variable on trivy scan jobs


festeveira:
It is currently impossible to configure TMPDIR environment variable on trivy scan jobs. This would be useful for clusters where hosts have limited storage capacity and downloading large images to emptyDir would not be optimal. Configuring TMPDIR allows users to mount a volume of their choice on that location.

**What steps did you take and what happened:**
Installed trivy-operator via helm chart and found it is impossible to configure TMPDIR on trivy scan jobs.

**What did you expect to happen:**
Possible to configure TMPDIR 

**Environment:**

- Trivy-Operator version (use `trivy-operator version`): 0.29.0
- Kubernetes version (use `kubectl version`): v1.32.7
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Debian 12

afdesk:
@festeveira thanks for your pruposal here!

I took a look at your PR and ideas.

Could you please describe your use case in a bit more detail, especially regarding why you need to customize TMPDIR?

To clarify — TMPDIR is primarily used by Trivy to store the local file cache and vulnerability database. By design, Trivy relies on a single-process bbolt database for these purposes, and there are currently no plans to change this behavior in the open-source version.

Because of this design choice, concurrent access to the same database directory (including through a shared TMPDIR) may lead to conflicts and failures.

If your goal is to have a shared cache across multiple instances, you can use Redis, which Trivy supports as a shared cache backend.

festeveira:
I believe I was mistaken. I observed the storage utilization in my kubernetes clusters go up on some nodes during scans causing evictions. Upon further investigation I found it was the scans on elastic-stack pods which have 3 elasticsearch containers. Upon reading these docs: https://trivy.dev/v0.64/docs/references/troubleshooting/, specifically:
`If the image is large or the temporary directory has insufficient space, the scan will fail. You can configure the directory path to redirect Trivy to a directory with adequate storage. On Unix systems, you can set the $TMPDIR environment variable.`, I believed that the image size was causing the issue, but no, it seems that it was the java-db download on all 3 containers in parallel which was causing the storage utilization to spike a few GBs. 

Now after testing with cacheDir pointing to a shared volume I see that it does not spike that much, only a few hundred MBs which is manageable. I also have `--slow` enabled.

This could still be an issue for users scanning very large images in parallel on the same host. I leave it to you to decide. Thank you for your time.

[...]

PR #2863, opened 2026-01-27
fix: use a different temporary directory for each scan

### LINKED ISSUE #2859: All scan containers are using the same temporary directory, which leads to failure

trivy temp directory should be unique. This is done by using the PID to discriminate. However, with trivy-operator pid is always 1 (one). The result is always the same: `/tmp/trivy-1`:

```golang
	tempDir := filepath.Join(os.TempDir(), fmt.Sprintf("trivy-%d", pid))
```
[ref](https://github.com/aquasecurity/trivy/blob/f97ac7e112a359138e7d33b3d0dc8f6da2166fd1/pkg/x/os/os.go#L25)

At the end of the scan, trivy is cleaning this directory ([ref](https://github.com/aquasecurity/trivy/blob/f97ac7e112a359138e7d33b3d0dc8f6da2166fd1/pkg/x/os/os.go#L87)). This can lead to `no such file or directory` when a file is deleted in a container while another container is about to delete it.

No failure of trivy scan jobs on multi-containers pods.

Proposed fix:  use a different emptyDir for each container (but keep `/tmp/trivy/.cache` and `/tmp/scan` shared).

**Environment:**

- Trivy-Operator version (use `trivy-operator version`):
- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

sathieu:
## Description

More info in #2859

## Related issues
- Close #2859

## Checklist

### DIFF
--- a/pkg/plugins/trivy/filesystem.go
+++ b/pkg/plugins/trivy/filesystem.go
@@ -70,19 +70,15 @@ func GetPodSpecForStandaloneFSMode(ctx trivyoperator.PluginContext, config Confi
 
 	cacheDir := config.GetFilesystemScanCacheDir()
 
-	volumeMounts := []corev1.VolumeMount{
-		{
-			Name:      FsSharedVolumeName,
-			ReadOnly:  false,
-			MountPath: "/var/trivyoperator",
-		},
-		{
-			Name:      tmpVolumeName,
-			MountPath: "/tmp",
-			ReadOnly:  false,
-		},
-	}
 	volumes := []corev1.Volume{
+		{
+			Name: trivyCacheVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					Medium: corev1.StorageMediumDefault,
+				},
+			},
+		},
 		{
 			Name: FsSharedVolumeName,
 			VolumeSource: corev1.VolumeSource{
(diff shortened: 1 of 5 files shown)


sathieu:
(marking as draft, as it's not completely working yet for all kind of scans filesystem/image/rootfs standalone/clientserver)

sathieu:
@afdesk Please tell me if this is the correct way to fix this. As I need to update a lot of tests and rebase (and fix conflict).

afdesk:
@sathieu thanks for your efforts and sorry for delay.
I had to think it over and checked one idea.

Initially, I assumed that enabling `cache-backed: memory` flag would eliminate all concurrency issues related to shared cache access. In practice, this does mitigate the problem in the majority of scenarios.

However, there are specific edge cases — particularly when dealing with large binary artifacts — where Trivy, by design, persists data to the system temporary directory. Under these conditions, the only reliable solution is to allocate a distinct temporary directory per scan to ensure proper isolation and avoid race conditions or file collisions.

Given this, I agree that we need a way to customize the tmp folder.

One additional clarification to prevent misinterpretation: instead of naming the directory `trivy-cache`, it would be more accurate to use something like `trivy-temp` or `trivy-tmp`. The directory in question is intended specifically for Trivy’s temporary artifacts rather than for cache storage, and the naming should reflect its actual purpose.

Thanks for your efforts!

Also there are a few related issues and PRs here:
* https://github.com/aquasecurity/trivy-operator/issues/2795
* https://github.com/aquasecurity/trivy-operator/pull/2796

sathieu:
This is no longer needed because of https://github.com/aquasecurity/trivy/commit/c8d1d0d113b0bd3a058f566ede786b18ff94f75e.

Closing.

PR #2901, opened 2026-02-27
docs: clarify potential flag collision

cruwe:
## Description

First, thank you very much for developing the trivy suite and making it available to the general public.

More specifically, accessGlobalSecretsAndServiceAccount may not be true for privateRegistryScanSecretsNames to be considered.

## Checklist


cruwe:
Thanks again for your continuing work on the trivy-operator.

I am politely requesting a review, as I consider my proposal to be still valid and have not noticed any attention. Please do not take my reminder as undue pressure.

Thanks

PR #2906, opened 2026-03-02
ci: pin all actions by SHA

simar7:
## Description

ci: pin all actions by SHA

Uses https://github.com/metcalfc/gh-refme

## Checklist

### DIFF
--- a/.github/workflows/title_checker-labeler.yaml
+++ b/.github/workflows/title_checker-labeler.yaml
@@ -13,7 +13,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5
+      - uses: amannn/action-semantic-pull-request@e32d7e603df1aa1ba07e981f2a23455dee596825 # was: amannn/action-semantic-pull-request@v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/title_checker-labeler.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5
+      - uses: amannn/action-semantic-pull-request@e32d7e603df1aa1ba07e981f2a23455dee596825 # was: amannn/action-semantic-pull-request@v5
```
afdesk, line 16: it's a bit strange, but the latest version v5.5.3 has another hash: https://github.com/amannn/action-semantic-pull-request/releases/tag/v5.5.3

simar7: Good catch - I feel this might be a bug then with the tool. The commit it referenced is here https://github.com/amannn/action-semantic-pull-request/commit/e32d7e603df1aa1ba07e981f2a23455dee596825

simar7: cc @nikpivkin

nikpivkin: The maintainer of this action did not rewrite the major tag to always point to the latest patch or minor version. Therefore, the `v5` tag is fixed to commit `e32d7e603df1aa1ba07e981f2a23455dee596825`.

image

simar7: thanks @nikpivkin - @afdesk since we were using v5, I think the SHA makes sense. WDYT?

afdesk: I agree

PR #2980, opened 2026-05-30
feat: distroless Trivy scan support, zstd wire format, streaming decoder

### LINKED ISSUE #2823: refactor: find a way to not rely on shell commands
Based on a comment https://github.com/aquasecurity/trivy-operator/pull/2725#discussion_r2536163667:

we should find a way to not rely on arbitrary shell commands within the operator.

bryantbiggs:
Closes #2823.

> [!IMPORTANT]
> **Depends on #2979.** That PR fixes the chart's silent dropping of `scanJobCompressLogs: false`. The new E2E scenario added here relies on the chart honouring overrides for the trivy image refs, which is independent — but the no-compress code path's user-facing override only works once #2979 lands. Merge #2979 first, then this.

## Summary

Three changes that together modernize the scan output pipeline so it (a) works on distroless Trivy images and (b) is materially faster and lighter than the prior bzip2-shell-pipe path:

1. **Distroless scan support.** The scan Job's container today is invoked as `/bin/sh -c "trivy ... 2>...; if [ \$rc -eq 1 ]; then cat …; else bzip2 -c … | base64; fi; exit \$rc"`. That assumes the Trivy image has a shell plus `cat`, `bzip2`, and `base64`. Chainguard's distroless Trivy image (`cgr.dev/chainguard/trivy`) has none of those, so scans fail today. This PR replaces the shell pipeline with two distroless-safe Command/Args shapes selected on `CompressLogs`:
   - `compressLogs=false` → `Command: [\"trivy\"]`, `Args: [..., \"--quiet\"]`. Trivy writes JSON straight to stdout.
   - `compressLogs=true` → `Command: [\"/var/trivyoperator/bin/scan-wrapper\"]`, where the wrapper is a new static Go binary (`cmd/scan-wrapper/`) shipped in the operator image and mounted into the scan Pod via an init container (only attached when CompressLogs=true).

2. **Wire format: bzip2 → zstd.** The on-the-wire compression switches from bzip2 to zstd. Same `base64` outer encoding (pod-log transport remains text-oriented). Smaller code, fewer direct deps, dramatically faster on the operator's decode hot path. Single library (`klauspost/compress/zstd`) handles both encode and decode — bzip2 required the awkward stdlib-reader + dsnet-writer split because Go stdlib has no bzip2 writer.

3. **Streaming decoder.** `pkg/utils.ReadCompressData` now streams base64 → zstd on the fly instead of materializing the full payload in memory twice (was: `io.ReadAll` → base64 decode → `bytes.Buffer` decompress). Constant memory for any report size. Signature also simplified to `func(io.Reader) io.ReadCloser` (drops the always-nil error return; loosens `io.ReadCloser` → `io.Reader` to match the documented caller-owned source contract).

## Benefits at a glance

| Aspect | Before | After |
|---|---|---|
| Scan container CLI deps | `/bin/sh`, `cat`, `bzip2`, `base64` | none (scan-wrapper handles everything) |
| Wire compression | bzip2 + base64 | zstd + base64 |
| Operator decode speed | bzip2 (baseline) | zstd, ~10–50× faster |
| Decode memory | full payload buffered twice | streaming, constant |
| Compress libs | stdlib `bzip2` (reader-only) + `dsnet/compress` (writer) | single `klauspost/compress/zstd` |
| Direct deps in `go.mod` | `dsnet/compress` direct | `dsnet/compress` demoted to indirect (kept by `trivy-kubernetes`); no new direct deps (`klauspost/compress` was already transitive) |
| `pkg/utils/compressutil.go` size | ~70 lines, two helpers + buffered pipeline | ~57 lines, single streaming pipeline |

## What's in this PR

| | |
|---|---|
| `cmd/scan-wrapper/` | new static Go binary; unit tests using the `os/exec` re-exec helper-process pattern; round-trip test against `pkg/utils.ReadCompressData` |
| `pkg/plugins/trivy/image.go` | refactored `getCommandAndArgs` + `GetSbomScanCommandAndArgs` to use a single `scanCommandAndArgs` helper that selects shape on CompressLogs; new `scanWrapperInstallContainer` + `scanWrapperVolumeAndMount` helpers |
| `pkg/plugins/trivy/image_test.go`, `plugin_test.go` | existing pod-spec fixtures updated to the new shapes; new `TestScanCommands_NoShellTokens` regression guard asserts no `/bin/sh`/`sh`/`bzip2`/`base64`/`cat` tokens in scan Command/Args under either compress mode; new `TestScanWrapperInstallContainer` tests the init-container wiring |
| `pkg/utils/compressutil.go` | swapped bzip2 → zstd; rewrote `ReadCompressData` as a streaming pipeline `bzip2/zstd(base64(skipWhitespace(bufio(src))))`; signature simplified to `(io.Reader) io.ReadCloser` |
| `pkg/plugins/trivy/plugin.go` | one call site of `ReadCompressData` updated for the new signature (drops an unreachable error check) |
| `tests/envtest/testdata/fixture/*-expected-scan.yaml` | regenerated against the controller's actual scan-Job output for the new scan-wrapper command shape |
| `build/trivy-operator/Dockerfile{,.ubi9,.fips.ubi9}` | `COPY scan-wrapper /usr/local/bin/scan-wrapper` |
| `.goreleaser.yaml`, `goreleaser-e2e.yaml` | new `scan-wrapper` builds entry (linux + same architectures as trivy-operator); bundled into every `dockers:` block |
| `deploy/helm/templates/deployment.yaml`, `deploy/static/trivy-operator.yaml` | `OPERATOR_IMAGE` env var on the operator Deployment |
| `tests/e2e/config/image-mode-distroless.yaml`, `tests/e2e/image-mode-distroless/workload/` | new E2E scenario: install the operator pointed at `cgr.dev/chainguard/trivy:latest` (used as the scanner runtime), scan an nginx pod end-to-end |
| `.github/workflows/build.yaml` | runs the new E2E scenario after the existing image-mode scenario, reusing the kind cluster |

## On dropping base64

Considered, kept. zstd output is binary and pod logs are a text-oriented transport (CRI runtime line framing on `\n`, varying behavior on Windows nodes, log-aggregator middleware in some clusters). The 33% overhead is bounded; the silent-corruption surface area of going raw-binary is not worth the savings on a one-shot per-scan payload.

## Test plan

- [x] `go test ./cmd/scan-wrapper/... ./pkg/utils/... ./pkg/plugins/trivy/...` — 247 pass locally with `GOEXPERIMENT=jsonv2`
- [x] envtest (`tests/envtest/`) passes against the regenerated scan-Job fixtures
- [x] `TestScanCommands_NoShellTokens` — fails loudly if `/bin/sh`/`sh`/`bzip2`/`base64`/`cat` ever return to scan Command/Args under either compress mode
- [x] Round-trip test: `scan-wrapper --compress`-encoded stdout decodes cleanly through the unchanged-shape `pkg/utils.ReadCompressData`
- [x] `pkg/utils/testdata/fixture/illegal_char_{before,after}.txt` regenerated as zstd; whitespace-tolerance test still passes
- [x] `helm template deploy/helm` renders `OPERATOR_IMAGE` env on the operator container
- [ ] CI \"Image scan mode against a distroless Trivy image (Chainguard)\" step passes end-to-end on kind
- [ ] All existing E2E scenarios still pass (compress=true via scan-wrapper exercising the same flow on a non-distroless image)

## Not in this PR (worth flagging)

- `OPERATOR_IMAGE` is read directly via `os.Getenv` from `pkg/plugins/trivy/image.go` rather than plumbed through `pkg/operator/etc.Config`. Keeps the diff small and matches the precedent of `OPERATOR_NAMESPACE`-style env vars on the chart; could move into `Config` as a follow-up if maintainers prefer.
- The new distroless E2E scenario covers `compressLogs=true` + distroless. A second scenario explicitly toggling `compressLogs=false` on the distroless image would prove the no-compress branch end-to-end as well, but is strictly weaker than what already runs.
- Rolling-upgrade compatibility: in-flight scan Jobs started under a bzip2-format operator and read by a zstd-format operator will fail to parse one report. The vulnerabilityreport controller already retries failed scans, so this self-heals in one reconcile cycle. No persistent state is affected (the format is internal to scan-Job stdout → pod log → operator).


bryantbiggs:
@eyalke / @afdesk I know its a rather large PR, mostly due to the fixture changes from swapping bzip2 to zstd, but I was wondering if we could at least chat through this PR and #3107 at a high level to start. the primary goal is to be able to use hardened/minimal images like Chainguard/minimus/distroless/etc without the need for shells, coreutils type tools, etc. The single binary would be ideal, but that would require the upstream trivy binary to support compression, so the next best approach is to bake as much of this compression functionality into a custom binary that can ship alongside the trivy binary and those two get built and copied into the minimal images without any other shells/tools required

Pinguladora:
any update on this? As Bryant said I find it quite relevant to avoid shells, given the increasing attacks to supply chain

PR #3038, opened 2026-08-17
fix(compliance): honour compliance.failEntriesLimit in the detail report

clain23:
### Description

`compliance.failEntriesLimit` is documented ("limit the number of fail entries per control check in the cluster compliance detail report — this limit is for preventing the report from being too large per control checks"), defaults to `10`, is rendered into the operator ConfigMap by the chart… and has **no call site**. `ConfigData.ComplianceFailEntriesLimit()` is only referenced by its own unit test:

```
pkg/trivyoperator/config.go:500:func (c ConfigData) ComplianceFailEntriesLimit() int {
pkg/trivyoperator/config_test.go:807:  gotLimit := tc.configData.ComplianceFailEntriesLimit()
```

`FromDetailReport` appends **every** failing misconfiguration of **every** control, so with `compliance.reportType: all` the `ClusterComplianceReport` grows with the number of workloads until the API server refuses to store it — and the report then silently stops updating, keeping stale data with a stale `updateTimestamp`:

```
"logger":"reconciler.clustercompliancereport","msg":"failed to generate compliance report",
"compliance report":{"name":"k8s-cis-1.23"},"error":"etcdserver: request is too large"

"controller":"clustercompliancereport","name":"k8s-nsa-1.0",
"error":"Request entity too large: limit is 3145728"
```

(the two shapes are the etcd `--max-request-bytes` cap and the API server's own body cap, depending on cluster configuration.)

Measured on a 9-node cluster with ~500 workloads, trivy-operator 0.33.0, `reportType: all`:

| | |
|---|---|
| `k8s-cis-1.23` `status` | **1 545 169 B** |
| control `5.7.3` alone | 697 217 B / **1241 checks** |
| same report with the configured limit of 10 applied | **125 079 B** |

So the report stops being writable at a fairly ordinary cluster size, and the setting that exists to prevent exactly that never runs.

### Change

* `FromDetailReport(sr, failEntriesLimit)` caps the failing checks kept per control. `<= 0` keeps the previous unlimited behaviour, so nothing changes for anyone who sets it to 0.
* `compliance.NewMgr` takes the `trivyoperator.ConfigData` (already available at the call site in `pkg/operator/operator.go`) and passes `ComplianceFailEntriesLimit()` through.
* A control with no misconfiguration still gets its synthetic passing check, whatever the limit.
* **Summary counts are unaffected** — `TotalsCheckCount(cr)` is computed from the full trivy result set before the detail report is built, so `passCount` / `failCount` stay exact. Only the per-control *sample* of failing targets is capped, which is what the setting documents.

### Tests

New `pkg/apis/aquasecurity/v1alpha1/compliance_types_test.go`: cap applied, below-cap untouched, `0` and negative meaning unlimited, `1` still recording the control as failing, and the passing-control case. Verified failing without the change and passing with it:

```
--- FAIL: TestFromDetailReport_FailEntriesLimit/caps_the_checks_kept_per_control
--- FAIL: TestFromDetailReport_FailEntriesLimit/limit_of_one_still_records_the_control_as_failing
```

`go build ./...`, `go vet` and `go test ./pkg/apis/aquasecurity/v1alpha1/... ./pkg/compliance/...` all green.

### Checklist

- [x] I've followed the [conventions](https://github.com/aquasecurity/trivy-operator/blob/main/CONTRIBUTING.md) in the PR title.
- [x] I've updated the documentation with the relevant information (the setting was already documented — this makes the documentation true).

### DIFF
--- a/pkg/apis/aquasecurity/v1alpha1/compliance_types.go
+++ b/pkg/apis/aquasecurity/v1alpha1/compliance_types.go
@@ -213,13 +213,20 @@ func FromSummaryReport(sr *report.SummaryReport) *SummaryReport {
 	}
 }
 
-// FromDetailReport map data from trivy summary report to crd summary report
-func FromDetailReport(sr *report.ComplianceReport) *ComplianceReport {
+// FromDetailReport map data from trivy summary report to crd summary report.
+// failEntriesLimit caps the number of failing checks kept per control — the
+// `compliance.failEntriesLimit` setting, which exists to keep the detail report
+// small enough to store. A value <= 0 means no limit.
+func FromDetailReport(sr *report.ComplianceReport, failEntriesLimit int) *ComplianceReport {
 	controlResults := make([]*ControlCheckResult, 0)
 	for _, sr := range sr.Results {
 		checks := make([]ComplianceCheck, 0)
+	collect:
 		for _, r := range sr.Results {
 			for _, ms := range r.Misconfigurations {
+				if failEntriesLimit > 0 && len(checks) >= failEntriesLimit {
+					break collect
+				}
 				checks = append(checks, ComplianceCheck{
 					ID:          ms.ID,
 					Target:      r.Target,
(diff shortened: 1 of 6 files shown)


clain23:
Closing this on our side — we're carrying the change in our own patched build instead and won't be maintaining it upstream. Sorry for the noise, and thanks for the project.

PR #3054, opened 2026-08-31
fix(configaudit): restore ConfigMap data for content-based…

LaibaBareera:
fix(configaudit): read ConfigMaps through uncached client

ConfigMap data was stripped from the shared informer cache, causing content-based config-audit policies to pass incorrectly.

Add `ConfigMap` to the manager client’s `DisableFor` configuration so ConfigMap reads go directly to the API server with complete `Data` and `BinaryData`. Continue stripping these fields from informer-cached objects to preserve the existing memory optimization.

Share the production client-cache configuration with envtest to prevent configuration drift. Add regression coverage using a local test policy ID and make the policy fixture self-contained with a private helper library.


pkg/configauditreport/controller/resource.go
```diff
@@ -53,6 +53,10 @@ type ResourceController struct {
 	ClusterVersion   string
 	CacheSyncTimeout time.Duration
 	ChecksLoader     *ChecksLoader
+	// APIReader reads straight from the API server, bypassing the shared
+	// informer cache. It is required to scan ConfigMap contents - see
+	// restoreConfigMapData.
+	APIReader client.Reader
```
nikpivkin: What about disabling the cache for `ConfigMap` instead of adding an `APIReader` field?

The manager already does this for `Secret` and `ServiceAccount`:

```go
DisableFor: []client.Object{
    &corev1.Secret{},
    &corev1.ServiceAccount{},
},
```

Adding `&corev1.ConfigMap{}` there makes every `ConfigMap` read through the manager client go to the API server. `ObjectResolver` then hands the reconciler a full object.

Reads of the two operator ConfigMaps go to the API server as well then. The main one is `pluginContext.GetConfig()` in `pkg/trivyoperator/plugin.go`, and it already does a live GET for the Secret in the same method, so it becomes two small GETs while a scan is being prepared. The policies `ConfigMap` is read rarely, `ChecksLoader` keeps them in memory.

The informer and the transform stay as they are, so memory does not change.

nikpivkin:
Unrelated to this PR, but I noticed it while reading the cache code. `Secret` is in `DisableFor`, but `TTLSecretReconciler` in `pkg/operator/ttl_report.go` still watches Secrets, so the informer keeps every `Secret` of the cluster in memory with its data, and the transform strips ConfigMaps only. Nothing ever reads those from the cache, since all Secret reads are live, and the reconciler itself only needs annotations and the creation timestamp, so a label selector in `cache.Options.ByObject` or a metadata only watch would drop that memory.

I can open an issue with the details if it is useful.

LaibaBareera:
@nikpivkin  You’re right. Disabling the cache for `ConfigMap` is a simpler approach and avoids adding a separate `APIReader` field. I’ve updated the implementation accordingly and fixed the related tests.

Good catch regarding the `Secret` informer as well. Since that is unrelated to this PR, opening a separate issue for it would be useful.

pkg/operator/cache.go
```diff
@@ -38,12 +36,8 @@ func CacheTransform() toolscache.TransformFunc {
...
 
 		if cm, ok := obj.(*corev1.ConfigMap); ok {
-			// Strip data from ALL ConfigMaps except the two operator ConfigMaps,
-			// which the operator reads through the cached client.
-			if cm.Name != trivyoperator.PoliciesConfigMapName && cm.Name != trivyoperator.TrivyConfigMapName {
-				cm.Data = nil
-				cm.BinaryData = nil
-			}
+			cm.Data = nil
+			cm.BinaryData = nil
```
afdesk, line 40: do we still need to strip the data?

LaibaBareera: Yes, we still need to strip the data. DisableFor only changes where reads through the manager client are served from; it does not prevent the ConfigMap informers from keeping objects in memory.

We still watch ConfigMaps in multiple places (ResourceController, ChecksLoader, and PolicyConfigController), so without the transform, the informer cache would retain their contents unnecessarily.

The two changes serve different purposes: DisableFor ensures reads that need ConfigMap contents go directly to the API server, while the transform prevents those contents from being retained in the informer cache.

afdesk (dismissed):
LGTM!

tests/envtest/testdata/fixture/policy.yaml
```diff
@@ -7,181 +7,6 @@ metadata:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
 data:
-  library.kubernetes.rego: |
-    package lib.kubernetes
```
nikpivkin, line 11: Why were the two library modules removed? `policy.1_host_ipc.rego` in the same ConfigMap still has `import data.lib.kubernetes`, and in the suite the operator namespace and the workload namespace are both default, so this ConfigMap is the real policies source there. The import now resolves from the built in bundle, and only while `trivy.useBuiltinRegoPolicies` is on. Please add one sentence about it to the description.

tests/envtest/testdata/content/policies/kubernetes/policies/general/configmap_with_secrets.rego
```diff
@@ -0,0 +1,45 @@
+# METADATA
+# title: "ConfigMap with secrets"
+# description: "Storing secrets in configMaps is unsafe."
+# scope: package
+# schemas:
+# - input: schema["kubernetes"]
+# custom:
+#   id: KSV109
```
nikpivkin: This rule is written from scratch and does not match the upstream `KSV109`, but it takes its `id` and `avd_id`. The spec then reads as coverage for the real check. A local id would avoid the confusion and the test still proves that ConfigMap data reaches Rego.

tests/envtest/suite_test.go
```diff
@@ -84,11 +92,29 @@ var _ = BeforeSuite(func() {
...
 
 	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
-		Scheme:     scheme.Scheme,
+		Scheme: scheme.Scheme,
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				// Mirror the operator's production client options, so that
+				// cache-related regressions - such as ConfigMap contents being
+				// stripped before config-audit reads them - are reproducible here.
+				DisableFor: []client.Object{
```
nikpivkin: The list is copied by hand here. `CacheTransform` was exported so the suite could not drift from production, the same applies to these client options. Please export them too, otherwise the next type added in `Start` will silently not be mirrored here.

nikpivkin (changes requested):
The description still says the fix reads ConfigMap contents through an uncached API reader during reconciliation, which is not what the code does now. Please update it, and the title too, since commits are squashed and this text ends up in main.

nikpivkin (approved):

afdesk (approved):

